PR #14, opened 2023-10-03
Update gevulot-overview.md

teempai:
Some small additions on fees, descriptions.

Started adding transaction types and info on the transaction flow through the network to communicate how it works better. If someone could check whether I've missed anything critical, that would be appreciated!


teempai:
Simplified proof transaction flow description.

tuommaki (approved):
LGTM :+1:

PR #24, opened 2024-01-10
Drop use of array_chunks feature from nightly

tuommaki:


crates/node/src/types/hash.rs
```diff
@@ -64,8 +64,17 @@ impl From<&[u8]> for Hash {
...
+        let mut offset = 0;
+        let mut chunks = vec![];
+        for _ in 0..value.0.len() / chunk_sz {
+            chunks.push(
+                <&[u8] as TryInto<[u8; 4]>>::try_into(&value.0[offset..offset + chunk_sz]).unwrap(),
+            );
+            offset += chunk_sz;
+        }
+
+        let mut chunks = chunks.iter_mut();
```
musitdev, line 77: Perhaps we can use direct array index because the Hash size is fixed and won't change like :
```
        let x = &value.0;
        Message(libsecp256k1::curve::Scalar([
            u32::from_be_bytes([x[0], x[1], x[2], x[3]]),
            u32::from_be_bytes([x[4], x[5], x[6], x[7]]),
            u32::from_be_bytes([x[8], x[9], x[10], x[11]]),
            u32::from_be_bytes([x[12], x[13], x[14], x[15]]),
            u32::from_be_bytes([x[16], x[17], x[18], x[19]]),
            u32::from_be_bytes([x[20], x[21], x[22], x[23]]),
            u32::from_be_bytes([x[24], x[25], x[26], x[27]]),
            u32::from_be_bytes([x[28], x[29], x[30], x[31]]),
        ]))
```

tuommaki: Ack, I went with this. Initially I wanted to go with chunking, because at least I haven't fully decided on the final hashing algorithm (nor the one for key) yet and hence wanted to maintain option to change it, but I just realized that this was incorrect thinking: The `Message` size is hardcoded anyway so there's no way around it that this won't change as-is :smile:

musitdev (approved):

crates/node/src/types/hash.rs
musitdev: I think about it later, but we can add a compile assert that verify that the Hash size is 32. So if we change the size, the compiler will tell it. I look for the code

musitdev: You can add this code in the from function to verify at compile time that the Hash size is 32. SO if we change it, it won't compile.
```
const _: [(); 0 - !{ const ASSERT: bool = ( HASH_SIZE == 32 ); ASSERT } as usize] = [];
```

musitdev: A simpler version. 
```
const _: [(); 0 - !{  ( HASH_SIZE == 32 ) } as usize] = [];
```

PR #26, opened 2024-01-10
Protoc in CI

tuommaki:


tuommaki:
YOLO; I'll fix clippy issues separately.

PR #33, opened 2024-01-12
Set limits to PgPool

tuommaki:
- Maximum number of connections (64).
- Connection timeout (750ms).

### DIFF
--- a/crates/node/src/storage/database/postgres.rs
+++ b/crates/node/src/storage/database/postgres.rs
@@ -1,9 +1,14 @@
+use std::time::Duration;
+
 use eyre::Result;
-use sqlx::{self, Row};
+use sqlx::{self, postgres::PgPoolOptions, Row};
 use uuid::Uuid;
 
 use super::entity::{self};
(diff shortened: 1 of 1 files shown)


crates/node/src/storage/database/postgres.rs
```diff
@@ -1,10 +1,15 @@
...
 use eyre::Result;
-use sqlx::{self, Row};
+use sqlx::{self, postgres::PgPoolOptions, Row};
 use uuid::Uuid;
 
 use super::entity::{self};
 use crate::types::{self, transaction::ProgramData, File, Hash, Program, Task};
 
+const MAX_DB_CONNS: u32 = 64;
+const DB_CONNECT_TIMEOUT: Duration = Duration::from_millis(750);
```
tuommaki, line 11: I'm a bit undecided whether or not these should be configurable.

musitdev: You'll decide when you'll need to configure it. It's easy to add some configuration and use these as default value when needed. If you don't have a good reason for now, let the things like that.

musitdev (approved):

PR #35, opened 2024-01-12
Restrict `proofs` table with UNIQUE(parent, prover)

tuommaki:
Currently `Program`s are unique within `Workflow` and therefore it's impossible to have same `parent` and `prover` hashes for multiple proofs.


crates/node/migrations/20231128120351_transactions-table.sql
```diff
@@ -79,6 +79,7 @@ CREATE TABLE proof (
     parent VARCHAR(64) NOT NULL,
     prover VARCHAR(64),
     proof BYTEA NOT NULL,
+    UNIQUE(parent, prover),
```
tuommaki, line 82: Adding seat belts when discovering pitfalls :grimacing:

musitdev (approved):

PR #39, opened 2024-01-17
Cli tool initial version

musitdev:
Add a cli tool:

Deploy prover/verifier
Exec program workflow.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,9 +4,10 @@ exclude = [
 ]
 members = [
   "crates/node",
+  "crates/cli",
   "crates/shim",
   "crates/shim-ffi",
   "crates/tests/test-programs",
-  "crates/tests/e2e-tests",
+  "crates/tests/e2e-tests", "crates/cli",
 ]
 resolver = "2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,6 @@ members = [
   "crates/shim",
   "crates/shim-ffi",
   "crates/tests/test-programs",
-  "crates/tests/e2e-tests", "crates/cli",
+  "crates/tests/e2e-tests",
 ]
 resolver = "2"
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -2,8 +2,7 @@
 name = "gevulot-cli"
 version = "0.1.0"
 edition = "2021"
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+license = "MIT OR Apache-2.0"
 
 [dependencies]
 anyhow = "1.0.71"
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -15,9 +15,12 @@ hex = "0.4"
 http-body-util = "0.1"
 hyper = { version = "1", features = ["full"] }
 hyper-util = { version = "0.1", features = ["full"] }
+indicatif = "0.17.7"
 libsecp256k1 = "0.7"
 log = "0.4.20"
 rand = { version = "0.8", features = [ "std_rng" ] }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = { version = "1.0"}
 sha3 = "0.10"
 tracing-subscriber = "0.3.18"
 tokio = { version = "1", features = [ "full" ] }
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "gevulot-cli"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.71"
+blake3 = { version = "1.5", features = [ "mmap" ] }
+clap = { version = "4.0.29", features = ["derive"] }
+futures-util = { version = "0.3", default-features = false }
+gevulot-node = { path = "../node" }
+hex = "0.4"
+http-body-util = "0.1"
+hyper = { version = "1", features = ["full"] }
+hyper-util = { version = "0.1", features = ["full"] }
+libsecp256k1 = "0.7"
+log = "0.4.20"
+rand = { version = "0.8", features = [ "std_rng" ] }
+sha3 = "0.10"
+tracing-subscriber = "0.3.18"
+tokio = { version = "1", features = [ "full" ] }
+tokio-util = "0.7"
+url = "2.5.0"
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -327,7 +327,7 @@ fn verify_file_path(file_path: &str) -> Option<PathBuf> {
         .and_then(|present| present.then_some(path))
 }
 
-//TODO add utillity function to Hash.
+// TODO: Add utility function to Hash.
 fn extract_hash_from_file_content(path: &PathBuf) -> Option<String> {
     let mut hasher = blake3::Hasher::new();
     let fd = std::fs::File::open(path).ok()?;
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -1,0 +1,334 @@
+use gevulot_node::{
+    rpc_client::RpcClient,
+    types::{
+        transaction::{Payload, ProgramData, ProgramMetadata, Workflow, WorkflowStep},
+        Hash, Transaction,
+    },
+};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::net::SocketAddr;
+use std::path::PathBuf;
+
+pub mod keyfile;
+mod server;
+
+type BoxResult<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+// {
+//     program: "Program Hash",
+//     cmd_args: [ {name: "args name", value:"args value"}, ...],
+//     inputs: vec![],
+// }
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct JsonCmdArgs {
+    name: String,
+    value: String,
+}
+
+impl From<JsonCmdArgs> for [String; 2] {
+    fn from(args: JsonCmdArgs) -> Self {
+        [args.name, args.value]
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct JsonExecArgs {
+    program: String,
+    cmd_args: Vec<JsonCmdArgs>,
+    inputs: Vec<JsonProgramData>,
+}
+
+impl TryFrom<JsonExecArgs> for WorkflowStep {
+    type Error = String;
+
+    fn try_from(data: JsonExecArgs) -> Result<Self, Self::Error> {
+        Ok(WorkflowStep {
+            program: (&(hex::decode(data.program)
+                .map_err(|err| format!("program decoding hash error:{err}"))?)[..])
+                .into(),
+            args: data
+                .cmd_args
+                .into_iter()
+                .flat_map(<[String; 2]>::from)
+                .collect(),
+            inputs: data
+                .inputs
+                .into_iter()
+                .map(|i| i.try_into())
+                .collect::<Result<Vec<ProgramData>, _>>()
+                .map_err(|err| format!("Source program decoding hash error:{err}"))?,
+        })
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub enum JsonProgramData {
+    Input {
+        file_name: String,
+        file_url: String,
+        checksum: String,
+    },
+    Output {
+        source_program: String,
+        file_name: String,
+    },
+}
+
+impl TryFrom<JsonProgramData> for ProgramData {
+    type Error = hex::FromHexError;
+
+    fn try_from(data: JsonProgramData) -> Result<Self, Self::Error> {
+        match data {
+            JsonProgramData::Input {
+                file_name,
+                file_url,
+                checksum,
+            } => Ok(ProgramData::Input {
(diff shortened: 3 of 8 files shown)


crates/cli/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+                        .and_then(|url| get_last_segment_from_url(url))
+                        .unwrap_or(tx_input_args.to_string());
+                    file_url_args
+                        .map(|img_url| (filename, img_url, tx_input_args))
+                        .ok_or("Image url not provided with Hash".to_string())
+                });
+            Some(res)
+        })
+        //never None
+        .unwrap()
```
tuommaki, line 313: This is a nice block! Reads very well! :heart_eyes:

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+fn get_last_segment_from_url(url: &str) -> Option<String> {
+    url::Url::parse(url).ok().and_then(|s| {
+        s.path_segments()
+            .and_then(|iter| iter.last().map(String::from))
+    })
+}
+
+//Verifiy that the path exist and it's a file.
```
tuommaki, line 323:
```suggestion
// Verify that the path exists and is a file.
```

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+fn verify_file_path(file_path: &str) -> Option<PathBuf> {
+    let path: PathBuf = file_path.into();
+    path.try_exists()
+        .map(|res| res && path.is_file())
+        .ok()
+        .and_then(|present| present.then_some(path))
+}
+
+//TODO add utillity function to Hash.
```
tuommaki, line 332:
```suggestion
// TODO: Add utility function to Hash.
```

crates/cli/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+#[clap(author = "Gevulot Team", version, about, long_about = None)]
+pub struct ArgConfiguration {
+    /// RPC url of the Gevulot node
+    #[clap(
+        short,
+        long,
+        default_value = "http://localhost:9944",
+        value_name = "URL"
+    )]
+    jsonurl: String,
```
tuommaki, line 17:
```suggestion
    json_url: String,
```
Similar to earlier comment on variable names, I feel like these params could also be formatted so that they are split to words. WDYT?

crates/cli/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+        prover: String,
+        /// file path containing the img of the verifier to deployor the hash  of the verifier img (--verifierimgurl is mandorty in this case). If the  file is not found the parameters is used as an hash.
+        #[clap(short, long, value_name = "VERIFIER FILE or HASH")]
+        verifier: String,
+        /// url to get the prover img. If provided the prover will use this url to get the prover img. If not the cli tool start a local HTTP server to server the file to the node.
+        #[clap(long, value_name = "PROVER URL")]
+        proverimgurl: Option<String>,
+        /// url to get the verifier img. If provided the verifier will use this url to get the verifier img. If not the cli tool start a local HTTP server to server the file to the node.
+        #[clap(long, value_name = "VERIFIER URL")]
+        verifierimgurl: Option<String>,
```
tuommaki, line 52:
```suggestion
        verifier_img_url: Option<String>,
```
Like above & elsewhere, WDYT?

crates/cli/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// Deploy prover and verifier.
+    GenerateKey,
+
+    /// Deploy prover and verifier.
+    #[command(arg_required_else_help = true)]
+    Deploy {
+        /// name of the deployment.
+        #[clap(short, long, value_name = "DEPLOYMENT NAME")]
+        name: String,
+        /// file path containing the img of the prover to deploy or the hash  of the prover img (--proverimgurl is mandorty in this case). If the  file is not found the parameters is used as an hash.
```
tuommaki, line 41:
```suggestion
        /// file path containing the program image of the prover to deploy or the hash  of the prover image file (--prover-img-url is mandatory in this case). If the  file doesn't exist, the parameter is used as a hash.
```

crates/cli/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+        #[clap(short, long, value_name = "PROVER FILE or HASH")]
+        prover: String,
+        /// file path containing the img of the verifier to deployor the hash  of the verifier img (--verifierimgurl is mandorty in this case). If the  file is not found the parameters is used as an hash.
```
tuommaki, line 44:
```suggestion
        /// file path containing the program image of the verifier to deploy or the hash  of the verifier image file (--verifier-img-url is mandatory in this case). If the file doesn't exist, the parameter is used as a hash.
```

crates/cli/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+        #[clap(short, long, value_name = "VERIFIER FILE or HASH")]
+        verifier: String,
+        /// url to get the prover img. If provided the prover will use this url to get the prover img. If not the cli tool start a local HTTP server to server the file to the node.
```
tuommaki, line 47:
```suggestion
        /// url to get the prover image. If provided the prover will use this URL to get the prover image file. If not the cli tool starts a local HTTP server to serve the file to the node.
```

crates/cli/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+        #[clap(long, value_name = "PROVER URL")]
+        proverimgurl: Option<String>,
+        /// url to get the verifier img. If provided the verifier will use this url to get the verifier img. If not the cli tool start a local HTTP server to server the file to the node.
```
tuommaki, line 50:
```suggestion
        /// url to get the verifier image. If provided the verifier will use this URL to get the verifier image. If not the cli tool starts a local HTTP server to serve the file to the node.
```

crates/cli/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    let client = RpcClient::new(args.jsonurl);
+
+    match args.command {
+        ConfCommands::GenerateKey => match gevulot_cli::keyfile::create_key_file(&args.keyfile) {
+            Ok(()) => println!(
+                "Key generated and saved in file:{}",
+                args.keyfile.to_str().unwrap_or("")
+            ),
+            Err(err) => println!("Error during key  file creation:{err}"),
```
tuommaki, line 110:
```suggestion
            Err(err) => println!("Error during key file creation:{err}"),
```

crates/cli/src/server.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+fn build_file_progress_bar(file_name: String, total_size: u64) -> ProgressBar {
+    let pb = ProgressBar::new(total_size);
+    pb.set_style(ProgressStyle::with_template("{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})")
+        .unwrap()
+        .with_key("eta", move |state: &ProgressState, w: &mut dyn Write| write!(w, "{}-{:.1}s", file_name, state.eta().as_secs_f64()).unwrap())
+        .progress_chars("#>-"));
+    pb
+}
```
tuommaki, line 208: :sunglasses:

tuommaki (commented):
Few spelling fix suggestions and opening discussion on couple style related matters we should probably align on in general.

Except for command argument structure, the design & code otherwise LGTM :+1:

Cargo.toml
```diff
@@ -4,9 +4,10 @@ exclude = [
 ]
 members = [
   "crates/node",
+  "crates/cli",
   "crates/shim",
   "crates/shim-ffi",
   "crates/tests/test-programs",
-  "crates/tests/e2e-tests",
+  "crates/tests/e2e-tests", "crates/cli",
```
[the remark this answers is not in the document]
musitdev: You prefer not to put the cli crate in the workplace.

crates/cli/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "gevulot-cli"
```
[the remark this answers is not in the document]
musitdev: For the name by exp I know I'm not very good, so I take what it's done in other BC. But you can define any name, I don't really have any good idea.
I agree, it's better if the tool have its own repo because update and release rhythm and their consequences are not the same for the node and the client.
More generally, I agree with the Rust idiomatic's way to define crate:  define small crate. The drawback is that it add complexity when you have a lot of crate not published in crates.io because you've to manage all the specific commit for all crates you use. It can become tedious when all crates move fast before the first release.

crates/cli/src/server.rs
musitdev: To be honest, I copy / paste an example related to file download in the repo.

Cargo.toml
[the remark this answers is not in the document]
tuommaki: It's already on line 7. Isn't this duplicate?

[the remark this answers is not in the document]
musitdev: Yes, I didn't see. Strange cargo didn't complain.

musitdev:
All check pass, I can merge?

tuommaki:
> All check pass, I can merge?

I think there are still pending typos open in the diff side or is my GitHub view somehow not showing the latest? Most notable there are couple `gelulot` vs. `gevulot` points. I think at least those would be good to fix before merge.

musitdev:
Yes, they were hidden by github. I didn't see. I thought that solved issue were hidden and not opened. I look at them.

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::net::SocketAddr;
+use std::path::PathBuf;
+
+pub mod keyfile;
+mod server;
+
+type BoxResult<T> = std::result::Result<T, Box<dyn std::error::Error>>;
```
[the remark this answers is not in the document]
musitdev: No `BoxResult` is not a convention. I define it to differentiate from the std::Result. I found very annoying that every result has the same name but are incompatible (not the same error).
I use this Box result to wrap an underlying lib error that I mostly wrap in a String error after to ease error result formatting for the user.
In a more general case, I prefer thiserror and define each specific error. Here to make it simple I use a String to describe the error.
I can change and use anyhow, but when you need to format the error in a simple and comprehensible message is can be tedious to manage the format of all the different type of error you get.
So to make it simple, I mostly use string error, keep the existing Box<dyn error>. I can spend some time to make it better or we can open an issue to get back on it.

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+        [args.name, args.value]
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct JsonExecArgs {
+    program: String,
+    cmd_args: Vec<JsonCmdArgs>,
+    inputs: Vec<JsonProgramData>,
+}
```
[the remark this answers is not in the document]
musitdev: I thought it was more user-friendly to express parameters as a key/valuer than a list of String.
If you invert some data you don't see, with name/value it's easier to see the error.

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+
+#[allow(clippy::too_many_arguments)]
+pub async fn run_deploy_command(
+    client: RpcClient,
+    keyfile: PathBuf,
+    name: String,
+    prover: String,
+    verifier: String,
+    proverimgurl: Option<String>,
+    verifierimgurl: Option<String>,
```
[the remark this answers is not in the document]
musitdev: Yes I agree. I didn't think. I'll do a find replace in the code.

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+        generate_tx_data(verified_file, tx_input_args, file_url_args, served_file_map)?;
+    let mut data = ProgramMetadata {
+        //TODO what is the role of the name?
+        name: image_file_name.clone(),
+        hash: Hash::default(),
+        image_file_name,
+        image_file_url: image_file_url.to_string(),
+        image_file_checksum,
+    };
+    //TODO use a new to be sure the update is done during creation.
```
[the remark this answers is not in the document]
musitdev: Yes, we can open an issue about that. There's the same issue for ProgramMetadata. You've to call update data to finish the build.

crates/cli/src/lib.rs
musitdev: Thanks, when you have result and option, it's often easier to use the functional style to manage all case in a clear way.

crates/cli/src/lib.rs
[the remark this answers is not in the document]
tuommaki: The `BoxResult` is all good as such. :+1:

crates/cli/src/lib.rs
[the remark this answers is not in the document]
tuommaki: https://github.com/gevulotnetwork/gevulot/issues/41

crates/cli/src/lib.rs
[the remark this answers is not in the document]
tuommaki: Let it be as-is for now, but I think we need to revisit this, because I'm not sure if flag arguments work correctly with this approach.

crates/cli/src/main.rs
musitdev: I use this for because - or _ in cli cmd parameters is not idiomatic, so I think we should use jsonurl or url for the paramters. I'll specify the parameter name in the [derive] and use a more idiomatic way to name Rust fields.

crates/cli/src/main.rs
musitdev: I'll do a find replace in the code and push the change.

PR #40, opened 2024-01-18
Add bootstrap node and peer list exchange and automatique connection

musitdev:
Add node peers list propagation.
When a node connects to another one, they exchange their peer list. If there's a missing connection, the node connects to the missing node. So one node connected to another, it gets all not connected nodes and connect to them.

To start, a bootstrap node must be specified. This part has not been changed. The node use the config to get one peer to connect to. When it connects, it gets all peer list and open a connection to them.

It's a raw version but seems to work.

Some edge cases are not completely managed. For example, if 2 nodes (2,3) connect at the same time to one node 1. Node 2 and 3 get the other address from node 1 and try to connect each other at the same time. So it opens 2 connections between both nodes, which is not allowed. So the lib close one. From my tests, only one is closed and not both.

Node disconnection is not completely tested. From my tests, to be notified of a disconnection, the disconnect function must be called. I'll do more test with real host to see how it works. Node disconnection remove unused nodes from the exchanged peer list.

I've tested using the unit test. I create 2 tests that seems to work locally.

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -22,6 +22,7 @@ hex = "0.4"
 home = "0.5"
 jsonrpsee = { version = "0.20", features = [ "client", "server" ] }
 libsecp256k1 = "0.7"
+log = "0.4.20"
 num-bigint = { version = "0.4", features = [ "serde" ] }
 num-traits = "0.2"
 parking_lot = "0.12"
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -22,7 +22,6 @@ hex = "0.4"
 home = "0.5"
 jsonrpsee = { version = "0.20", features = [ "client", "server" ] }
 libsecp256k1 = "0.7"
-log = "0.4.20"
 num-bigint = { version = "0.4", features = [ "serde" ] }
 num-traits = "0.2"
 parking_lot = "0.12"
--- a/crates/node/src/networking/p2p.rs
+++ b/crates/node/src/networking/p2p.rs
@@ -441,8 +441,8 @@ mod tests {
 
         tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
 
-        assert_eq!((&peer1.peer_list.read()).len(), 1);
-        assert!((&peer1.peer_addr_mapping.read()).is_empty());
+        assert_eq!(peer1.peer_list.read().len(), 1);
+        assert!(peer1.peer_addr_mapping.read().is_empty());
     }
 
     #[tokio::test]
--- a/crates/node/src/networking/p2p.rs
+++ b/crates/node/src/networking/p2p.rs
@@ -218,10 +218,6 @@ impl Handshake for P2P {
         //do peer comparition
         let mut local_diff = {
             let local_peer_list: &mut BTreeSet<SocketAddr> = &mut self.peer_list.write();
-            let distant_diff: BTreeSet<SocketAddr> = local_peer_list
-                .difference(&distant_peer_list)
-                .cloned()
-                .collect();
             let local_diff: BTreeSet<SocketAddr> = distant_peer_list
                 .difference(local_peer_list)
                 .cloned()
@@ -356,10 +352,6 @@ mod tests {
             .unwrap();
 
         tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
-
-        //assert_eq!(peer1.get_connected_peers().len(), 3);
-        //assert_eq!(peer2.get_connected_peers().len(), 3);
-        //assert_eq!(peer3.get_connected_peers().len(), 3);
 
         tracing::debug!("send tx from peer2 to peer1 and peer3");
         let tx = new_tx();
--- a/crates/node/src/networking/p2p.rs
+++ b/crates/node/src/networking/p2p.rs
@@ -285,9 +285,9 @@ impl OnDisconnect for P2P {
 mod tests {
     use super::*;
     use eyre::Result;
-    use gevulot_node::types::{transaction::Payload, Hash, Transaction};
+    use gevulot_node::types::{transaction::Payload, Transaction};
     use libsecp256k1::SecretKey;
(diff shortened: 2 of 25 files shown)


crates/node/Cargo.toml
```diff
@@ -22,6 +22,7 @@ hex = "0.4"
 home = "0.5"
 jsonrpsee = { version = "0.20", features = [ "client", "server" ] }
 libsecp256k1 = "0.7"
+log = "0.4.20"
```
tuommaki, line 25: Accidental leftover?

crates/node/src/networking/p2p.rs
```diff
@@ -175,9 +313,141 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn test_peer_list_inter_connection() {
+        //start_logger(LevelFilter::ERROR);
```
tuommaki, line 318:
```suggestion
        start_logger(LevelFilter::ERROR);
```
In the merged version, I'd prefer debug logs silenced, even if the normal `cargo test` hides the output. WDYT?

crates/node/src/networking/p2p.rs
```diff
@@ -175,9 +313,141 @@ mod tests {
...
+            .node()
+            .connect(peer1.node().listening_addr().unwrap())
+            .await
+            .unwrap();
+
+        tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+
+        //assert_eq!(peer1.get_connected_peers().len(), 3);
+        //assert_eq!(peer2.get_connected_peers().len(), 3);
+        //assert_eq!(peer3.get_connected_peers().len(), 3);
```
tuommaki, line 362: Should these be uncommented? If not, maybe remove them?

crates/node/src/networking/p2p.rs
```diff
@@ -175,9 +313,141 @@ mod tests {
...
+        let recv_tx = rx1.recv().await.expect("sink recv");
+        assert_eq!(tx, recv_tx);
+        tracing::debug!("recv tx on peer2 from peer3");
+        let recv_tx = rx2.recv().await.expect("sink recv");
+        assert_eq!(tx, recv_tx);
+    }
+
+    #[tokio::test]
+    async fn test_two_peers_disconnect() {
+        //start_logger(LevelFilter::ERROR);
```
tuommaki, line 387:
```suggestion
        start_logger(LevelFilter::ERROR);
```

crates/node/src/networking/p2p.rs
```diff
@@ -175,9 +313,141 @@ mod tests {
...
+        tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+
+        assert_eq!(peer1.peer_list.read().len(), 1);
+        assert!(peer1.peer_addr_mapping.read().is_empty());
+    }
+
     #[tokio::test]
     async fn test_two_peers() {
-        start_logger(LevelFilter::ERROR);
+        //start_logger(LevelFilter::ERROR);
```
tuommaki, line 450:
```suggestion
        start_logger(LevelFilter::ERROR);
```

crates/node/src/networking/p2p.rs
```diff
@@ -111,6 +124,122 @@ impl Handshake for P2P {
...
+            }
+        };
+
+        //do peer comparition
+        let mut local_diff = {
+            let local_peer_list: &mut BTreeSet<SocketAddr> = &mut self.peer_list.write();
+            let distant_diff: BTreeSet<SocketAddr> = local_peer_list
+                .difference(&distant_peer_list)
+                .cloned()
+                .collect();
```
tuommaki, line 224:
```suggestion
```
For all I can read & test, this seems to be unused?

I wonder why the compiler / linters don't highlight this?

tuommaki (commented):
Some minor notes. In general LGTM :slightly_smiling_face:

crates/node/src/networking/p2p.rs
musitdev: I remove it because in the CI you don't need it and as all test load it if you start all the test in //  it breaks because it's already loaded by another test.
You run the test one after the other in the CI ?

crates/node/src/networking/p2p.rs
musitdev: Yes it's not use and strange it's not detected. I'll remove it

crates/node/src/networking/p2p.rs
musitdev: Yes, I'll remove them.

crates/node/src/networking/p2p.rs
tuommaki: > I remove it because in the CI you don't need it and as all test load it if you start all the test in // it breaks because it's already loaded by another test.

Ah! I wasn't aware of this. Zap it away then :+1:

> You run the test one after the other in the CI ?

The CI runs the tests with the default settings of `cargo test`.

musitdev: Strange if I run cargo test locally I have this error: 
```
Unable to install global subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```
I've to remove 2 from 3 to pass the test. 
tracing_subscriber::init() must be call only one time per execution. Perhaps the CI start an exe for each test.
But local execution break with the logger inited for each test.
I'm not aware of the last dev, but I think that the default Rust unit test executor don't have per or post execution for all test.
You really need login for the test?

tuommaki: > You really need login for the test?

No, not until debugging the test runs :slightly_smiling_face: 

> Zap it away then
By this I mean: Go ahead and delete the logging setup calls. Shouldn't that fix it?

musitdev: Usually I left the comment so that when you need to debug it's easier to activate.

tuommaki (approved):
Besides the `log` dependency, this LGTM :+1:

crates/node/Cargo.toml
musitdev: Yes I miss it. Removed.

PR #42, opened 2024-01-19
Add `Containerfile` to build node container

tuommaki:
The first version is based on Debian, but that choice shall be revisited later to optimize the container size.


tuommaki:
@musitdev added you as a reviewer to get notification on this, I'll merge this without wait though, to move fast for now.

PR #43, opened 2024-01-21
Implement transaction authorization

tuommaki:
Gevulot devnet is permissioned due to its nature. For this need,
implement basic transaction authorization based on authors' public keys.

### DIFF
--- a/crates/node/migrations/20231128120351_transactions-table.sql
+++ b/crates/node/migrations/20231128120351_transactions-table.sql
@@ -12,6 +12,7 @@ DROP TYPE IF EXISTS transaction_kind;
 CREATE TYPE transaction_kind AS ENUM ('empty','transfer', 'stake', 'unstake', 'deploy', 'run', 'proof', 'proofkey', 'verification', 'cancel');
 
 CREATE TABLE transaction (
+    author VARCHAR(130) NOT NULL,
     hash VARCHAR(64) PRIMARY KEY NOT NULL,
     kind transaction_kind NOT NULL,
     nonce NUMERIC NOT NULL,
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -16,7 +16,7 @@ use clap::Parser;
 use cli::{Cli, Command, Config, GenerateCommand, NodeKeyOptions, PeerCommand};
 use eyre::Result;
 use gevulot_node::types;
-use libsecp256k1::SecretKey;
+use libsecp256k1::{PublicKey, SecretKey};
 use pea2pea::Pea2Pea;
 use rand::{rngs::StdRng, SeedableRng};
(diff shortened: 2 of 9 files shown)


crates/node/migrations/20231128120351_transactions-table.sql
```diff
@@ -12,6 +12,7 @@ DROP TYPE IF EXISTS transaction_kind;
 CREATE TYPE transaction_kind AS ENUM ('empty','transfer', 'stake', 'unstake', 'deploy', 'run', 'proof', 'proofkey', 'verification', 'cancel');
 
 CREATE TABLE transaction (
+    author VARCHAR(130) NOT NULL,
```
tuommaki, line 15: This is a bit drive-by change and I wouldn't normally combine these, but gotta move fast :see_no_evil:

crates/node/src/main.rs
```diff
@@ -140,22 +153,20 @@ impl workflow::TransactionStore for storage::Database {
     }
 }
 
-struct AuthenticatingTxHandler {
```
tuommaki, line 143: Authorizing in `mempool` is better approach than in P2P handler. `mempool` captures both, JSON-RPC & P2P.

crates/node/src/main.rs
```diff
@@ -76,14 +76,27 @@ async fn main() -> Result<()> {
...
+                db.acl_deny(&key).await
             }
         },
         Command::Run { config } => run(Arc::new(config)).await,
+        Command::Show { op } => match op {
+            ShowCommand::PublicKey { key_file } => {
+                let bs = std::fs::read(key_file)?;
+                let key = SecretKey::parse(bs.as_slice().try_into()?)?;
+                let public_key = PublicKey::from_secret_key(&key);
+                println!("{}", hex::encode(public_key.serialize()));
```
musitdev, line 96: Use log::trace! instead of println!

musitdev (approved):
It opens a lot of discussion about software architecture, auth needs, ... but as you say we've to move fast. So we'll see later.

crates/node/src/main.rs
tuommaki: Actually, isn't `println!` totally fine in this case as it's purely meant to be "write to stdout" for end user, no matter what the logging configuration is?

tuommaki:
> It opens a lot of discussion about software architecture, auth needs, ... but as you say we've to move fast. So we'll see later.

True, but to make it even funnier, once we get into testnet phase, the authorization goes away :stuck_out_tongue_closed_eyes:

crates/node/src/main.rs
musitdev: Ah ok, I didn't really investigate the reason, I thought it was one forgets.

PR #44, opened 2024-01-21
Persist tx execution state

tuommaki:
In order to avoid re-executing old transactions on every restart, maintain execution state.


musitdev (approved):
I think that `propagated` and `executed` are not really part of the Tx, but a more state that indicate at what stage they are in their processing pipeline. For that type-state programming works well in Rust.
It's another thing that we'll discuss when thing will be more calm.

tuommaki:
> I think that propagated and executed are not really part of the Tx, but a more state that indicate at what stage they are in their processing pipeline. For that type-state programming works well in Rust.
> It's another thing that we'll discuss when thing will be more calm.

:100: 

I totally agree with you. As soon as we get this thing in the air, let's refactor this into better direction.

PR #45, opened 2024-01-21
Prepare for mainlining first release devnet node

tuommaki:
This commit is a next step towards Gevulot devnet.

Old test provers are removed. README is changed towards node-centric repository structure and specific INSTALL guide has been added.


INSTALL.md
tuommaki, line 1: @teempai Here's the installation guide. WDYT?

tuommaki:
I'll merge this now and fold the `proto` -> `main` shortly afterwards. Let's iterate & make frequent fixes when we come up with points to change. We can still discuss under this PR for a short while afterwards as well if it makes sense (i.e. there's suitable context).

PR #47, opened 2024-01-22
Add P2P beacon mode

tuommaki:
P2P beacon mode allows for lightweight node instance to run, with only purpose to coordinate other nodes' peer discovery. This is effectively a special node in the network that bridges others.

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -136,6 +136,25 @@ pub enum PeerCommand {
     },
 }
 
+#[derive(Debug, Args)]
+pub struct P2PBeaconConfig {
+    #[arg(
+        long,
+        long_help = "P2P listen address",
+        env = "GEVULOT_P2P_LISTEN_ADDR",
+        default_value = "127.0.0.1:9999"
+    )]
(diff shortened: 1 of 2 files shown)


musitdev:
I think about it during the meeting. The issue is that the download manager use the peer list to know other nodes IP to download the Tx files.
So these nodes won't have HTTP port open that can lead to connection timeout and slow the Tx file download process.
I think the best is to merge this PR in main so that I can get it and update the process to add the beacon node info in the peer list and update my download process accordingly.

tuommaki:
> So these nodes won't have HTTP port open that can lead to connection timeout and slow the Tx file download process.

When the node is not open, the connection attempt fails immediately (to `connection refused`). The slow-down should be rather minimal, I think?

> I think the best is to merge this PR in main so that I can get it and update the process to add the beacon node info in the peer list and update my download process accordingly.

Yup, I'll do this, but gotta do couple tests first :slightly_smiling_face:

tuommaki:
This is now running. Let's test this from different angles and iterate as we find issues.

PR #49, opened 2024-01-23
Don't broadcast propagated transaction

tuommaki:
If transaction has already been propagated (tx.propagated == true), don't send it to P2P network anymore.

### DIFF
--- a/crates/node/src/mempool/mod.rs
+++ b/crates/node/src/mempool/mod.rs
@@ -80,12 +80,14 @@ impl Mempool {
         self.storage.set(&tx).await?;
 
         // Broadcast new transaction to P2P network if it's configured.
-        if let Some(ref tx_chan) = self.tx_chan {
-            if tx_chan.send_tx(&tx).await.is_ok() {
-                tx.propagated = true;
-                self.storage.set(&tx).await?;
-            } else {
-                // TODO: Implement retry?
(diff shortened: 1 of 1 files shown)


musitdev (approved):
Just one remarks. tx_chan.send_tx(&tx) use p2p broadcast that doesn't fail if one node fail to receive the Tx. It only fails if there's no protocol writing registered.
Do we have to manage a fail when the Tx is send?

tuommaki:
> Just one remarks. tx_chan.send_tx(&tx) use p2p broadcast that doesn't fail if one node fail to receive the Tx. It only fails if there's no protocol writing registered.
> Do we have to manage a fail when the Tx is send?

Sounds like it wouldn't be easy in the current setting? It could also be a bit pointless, as there's no "catch-up replication" implemented, where a node would pull missing data if it was offline for a period of time. I'm not really sure if any of these are necessary during the devnet period?

musitdev:
If the node crash or restart, it should bootstrap from another node to get all missing data. The fact is that an error can occur during sending the Tx without because the distant node has crashed. Some heartbeat should be implemented if not. If the heartbeat fails, the node that has network issue should restart and bootstrap. If the connection issue last less than the heartbeat, the node should resent the tx to avoid restarting both node.

PR #52, opened 2024-01-23
Fix returned workspace volume path

tuommaki:
The current implementation to manage workspace volumes is not very ideal for multiple reasons:
- Outsourcing the job to external Go program is not very efficient.
- We lack substantial amount of control over the volume details.
  - ...which leads to current state that it's horrible fragile hack in many ways.

I have started writing the Rust implementation of TFS, but it's half-way and there are several dragons due to fact that there's no documentation and it's fully written based on the existing Go code.


tuommaki:
Due to order or various events, this issue wasn't discovered while testing the earlier workspace volume handling fix. Now it works though, so I'm gonna fast-forward here and directly merge.

musitdev (approved):

PR #56, opened 2024-01-24
Fallback to `PodmanArgs` escape hatch on ulimits

tuommaki:
On older podman instances, especially on older Ubuntu installations, there's no support for `Ulimit` configuration directive. To allow for same behaviour, use `PodmanArgs`.


tuommaki:

PR #58, opened 2024-01-24
Tx file propagation

musitdev:
Add Deploy Tx prover and image file propagation among all nodes.

Does a Run Tx can contain a download file? I'm not sure, it works in this case because of the path. The download manager server its file from the image directory. 

I correct an issue with the Tx propagation. The Tx wasn't processed the same way if it came from RPC or p2p. The correction is hard coded to avoid to add another Mutex.


crates/cli/localkey.pki
tuommaki, line 1: I'd maybe drop this from the repo & add a `.gitignore` entry for `localkey.pki` or `*.pki`.

crates/cli/src/keyfile.rs
```diff
@@ -21,3 +23,5 @@ pub fn read_key_file(file_path: &PathBuf) -> crate::BoxResult<SecretKey> {
     let key_array = fs::read(file_path)?;
     Ok(SecretKey::parse_slice(&key_array)?)
 }
+
+//pub print_pub_key
```
tuommaki, line 27:
```suggestion
```
TODO or just a leftover?

crates/node/src/cli.rs
```diff
@@ -44,12 +44,7 @@ pub struct Config {
     )]
     pub node_key_file: PathBuf,
 
-    #[arg(
-        long,
-        long_help = "",
-        env = "GEVULOT_P2P_DISCOVERY_ADDR",
-        default_value = "34.88.251.176:9999"
-    )]
+    #[arg(long, long_help = "", env = "GEVULOT_P2P_DISCOVERY_ADDR")]
```
tuommaki, line 47: Why drop the default value?

crates/node/src/main.rs
```diff
@@ -296,9 +310,13 @@ async fn run(config: Arc<Config>) -> Result<()> {
     .await?;
 
     tracing::info!("gevulot node started");
-    loop {
-        sleep(Duration::from_secs(1));
+    // loop {
+    //     sleep(Duration::from_secs(1));
+    // }
```
tuommaki, line 315:
```suggestion
```

I guess we can drop this?

crates/node/src/main.rs
```diff
@@ -296,9 +310,13 @@ async fn run(config: Arc<Config>) -> Result<()> {
...
+    if let Err(err) = download_jh.await {
+        tracing::info!("dwn mng error:{err}");
```
tuommaki, line 317:
```suggestion
        tracing::info!("download_manager error:{err}");
```
Less confusing if suddenly confronted with this one :sweat_smile:

crates/node/src/networking/download_manager.rs
```diff
@@ -1,21 +1,104 @@
...
+use std::path::Path;
+use tokio::fs::File;
+use tokio::net::TcpListener;
+use tokio::task::JoinHandle;
+use tokio_util::io::ReaderStream;
 
-use crate::{cli::Config, storage};
+// pub struct DownloadManager {
+//     file_server_jh: JoinHandle<()>,
+// }
```
tuommaki, line 19:
```suggestion
```

Let's drop this. I guess it was dead code anyway :see_no_evil:

crates/node/src/networking/download_manager.rs
```diff
@@ -1,21 +1,104 @@
...
 
-pub struct DownloadManager {
-    database: Arc<storage::Database>,
-    file_storage: Arc<storage::File>,
+// impl DownloadManager {
+//     pub async fn new(config: &Config, bind_addr: SocketAddr) -> Result<Self> {
+//         let file_server_jh = serve_file(config).await?;
+//         Ok(DownloadManager { file_server_jh })
+//     }
+// }
```
tuommaki, line 26:
```suggestion
```

Same here.

crates/node/src/networking/mod.rs
```diff
@@ -1,8 +1,8 @@
-mod download_manager;
+pub mod download_manager;
 mod noise;
 pub mod p2p;
 
-#[allow(unused_imports)]
-pub use download_manager::DownloadManager;
+// #[allow(unused_imports)]
+// pub use download_manager::DownloadManager;
```
tuommaki, line 6:
```suggestion
```

tuommaki (commented):
Check the default value of P2P discovery address. Was that intentional?

Otherwise, minor nitpicks/suggestions; in general LGTM :+1:

crates/cli/localkey.pki
musitdev: Yes good idea I add it. I didn't see I've added it.

crates/node/src/cli.rs
musitdev: To manage the case the node start without connecting, sort of primary bootstraper. In this case, you specify nothing and the node didn't try to connect.

tuommaki (approved):
LGTM :+1:

tuommaki:

PR #60, opened 2024-01-25
Add P2P bootstrap configuration to INSTALL.md

tuommaki:
Default value for P2P bootstrap node is going to be removed in another PR so it must be always configured explicitly.


tuommaki:

PR #63, opened 2024-01-26
Asset hash verification + file download propagation for input parameters

musitdev:
Update file propagation download for all files: 
 * deploy Tx
 * Run Tx input parameters
 
Use the same logic for all cases.

Add file checksum verification during download.

Test with 2 nodes.

### DIFF
--- a/crates/node/src/asset_manager/mod.rs
+++ b/crates/node/src/asset_manager/mod.rs
@@ -124,9 +124,6 @@ impl AssetManager {
     }
 
     async fn process_run(&self, tx: &Transaction) -> Result<()> {
-        // TODO: Process and program files are now downloaded in different ways. Combine these.
-        //let file_storage = storage::File::new(&self.config.data_directory);
-
         let workflow = match tx.payload.clone() {
             Payload::Run { workflow } => workflow,
             _ => return Err(AssetManagerError::IncompatibleTxPayload(tx.hash).into()),
--- a/crates/node/src/asset_manager/mod.rs
+++ b/crates/node/src/asset_manager/mod.rs
@@ -1,3 +1,11 @@
+use crate::{
+    cli::Config,
+    storage::Database,
+    types::{
+        transaction::{self, Transaction},
+        Hash, Program,
+    },
+};
 use eyre::{eyre, Result};
 use gevulot_node::types::{
     self,
@@ -7,16 +15,7 @@ use std::collections::HashMap;
 use std::net::SocketAddr;
 use std::{path::PathBuf, sync::Arc, time::Duration};
 use thiserror::Error;
-use tokio::{io::AsyncWriteExt, time::sleep};
-
-use crate::{
-    cli::Config,
-    storage::{self, Database},
-    types::{
-        transaction::{self, Transaction},
-        Hash, Program,
-    },
-};
+use tokio::time::sleep;
 
 #[allow(clippy::enum_variant_names)]
 #[derive(Error, Debug)]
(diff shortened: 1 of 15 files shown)


crates/node/src/asset_manager/mod.rs
```diff
@@ -115,21 +114,16 @@ impl AssetManager {
...
-            .join(program.image_file_name.clone());
-
-        let url = reqwest::Url::parse(&program.image_file_url)?;
-        self.download(&url, &file_path).await
+        self.download_image(
+            &program.image_file_url,
+            program.hash,
+            &program.image_file_name,
+            &program.image_file_checksum,
+        )
```
tuommaki, line 122: This function is always used just for programs. Maybe change the function param to take `&Program` ?

crates/node/src/networking/download_manager.rs
```diff
@@ -8,12 +9,101 @@ use hyper::server::conn::http1;
...
+            }
+            match resp {
+                Some(resp) => resp,
+                _ => {
+                    return Err(eyre!(
+                        "Download no host found to download the file: {:?}",
+                        file
+                    ));
+                }
+            }
```
tuommaki, line 62:
```suggestion
            resp.ok_or(Err(eyre!("Download no host found to download the file: {:?}", file)));
```
Could that work? I didn't check the `cargo fmt` or so, but would that make it a bit nicer?

crates/node/src/networking/download_manager.rs
```diff
@@ -8,12 +9,101 @@ use hyper::server::conn::http1;
...
+
+        while let Some(chunk) = resp.chunk().await? {
+            hasher.update(&chunk);
+            fd.write_all(&chunk).await?;
+        }
+
+        fd.flush().await?;
+        let checksum: gevulot_node::types::Hash = (&hasher.finalize()).into();
+        if checksum != file_hash {
+            Err(eyre!("Download file: {:?}, bad checksum", file))
```
tuommaki, line 93: The downloaded file is left on disk even if it failed the checksum. Should it be deleted or should just account for it in some kind of garbage collection if we implement such in the future?

crates/node/src/networking/download_manager.rs
```diff
@@ -8,12 +9,101 @@ use hyper::server::conn::http1;
...
 use hyper_util::rt::TokioIo;
+use std::net::SocketAddr;
 use std::path::Path;
 use tokio::fs::File;
+use tokio::io::AsyncWriteExt;
 use tokio::net::TcpListener;
 use tokio::task::JoinHandle;
 use tokio_util::io::ReaderStream;
 
+/// download downloads file from the given `url` and saves it to file in `file_path`.
```
tuommaki, line 20:
```suggestion
/// download_file downloads file from the given `url` and saves it to file in `local_directory_path` + / + `file`.
```

tuommaki (approved):
Couple notes, LGTM :+1:

crates/node/src/asset_manager/mod.rs
musitdev: Yes and download image will always come from a program.

crates/node/src/networking/download_manager.rs
musitdev: Yes, the file should be removed, and the program removed or inactivate too because currently, if the error arrives that download will always restart because the program is written in the db and the scheduler try to restart it.
I remove the program too or inactivate some way the program.

PR #65, opened 2024-01-29
Refactor P2P messaging

tuommaki:
- Support use cases where P2P node cannot be directly contacted. -> Allow configuration of advertised P2P address.
- Introduce versioned message types both for handshake and actual communication.
- Introduce support for node diagnostics in P2P network.


crates/node/build.rs
```diff
@@ -1,4 +1,9 @@
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    vergen::EmitBuilder::builder()
+        .build_timestamp()
+        .all_git()
+        .emit()?;
```
tuommaki, line 5: Capture information from build so that we can identify which versions of the node are running in the network.

crates/node/src/main.rs
```diff
@@ -201,10 +201,11 @@ async fn run(config: Arc<Config>) -> Result<()> {
 
     let p2p = Arc::new(
         networking::P2P::new(
-            "mempool-pubsub",
+            "gevulot-p2p-network",
```
tuommaki, line 204: It's not only mempool-pubsub anymore :slightly_smiling_face:

crates/node/src/mempool/mod.rs
```diff
@@ -108,12 +106,3 @@ impl networking::p2p::TxHandler for P2PTxHandler {
...
 }
-
-#[async_trait::async_trait]
-impl networking::p2p::TxChannel for networking::p2p::P2P {
-    async fn send_tx(&self, tx: &Transaction) -> Result<()> {
-        let bs = Bytes::from(bincode::serialize(tx)?);
-        self.broadcast(bs)?;
-        Ok(())
-    }
-}
```
tuommaki, line 119: I have zero idea why I had put this here but it's in the wrong place :see_no_evil:

crates/node/src/networking/p2p/protocol.rs
```diff
@@ -0,0 +1,43 @@
+use std::{collections::BTreeSet, net::SocketAddr};
+
+use gevulot_node::types;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) enum Handshake {
+    V1(HandshakeV1),
```
tuommaki, line 8: Later on `V2(HandshakeV2)` etc.

crates/node/src/networking/p2p/protocol.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) struct HandshakeV1 {
+    pub my_p2p_listen_addr: SocketAddr,
+    pub peers: BTreeSet<SocketAddr>,
+    pub http_port: Option<u16>,
+}
```
tuommaki, line 16: Having ability to describe the protocol message improves quality of life several orders of magnitude, compared to working with individual values directly written into pipe. :relieved:

crates/node/src/networking/p2p/protocol.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) enum Message {
+    V0(MessageV0),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) enum MessageV0 {
+    Transaction(types::Transaction),
+    DiagnosticsRequest(DiagnosticsRequestKind),
+    DiagnosticsResponse(DiagnosticsResponseV0),
```
tuommaki, line 27: Normally I'd put these into separate PR, but I bundled them here in order to demonstrate the thinking with these type definitions.

Normally, for now, there are only `Transaction()` kinds broadcasted in the network. No need for (request, response) pair since those aren't ever replied.

The diagnostics messages are different, however, because there the use case is such that one client can send a request where others are then expected to reply.

crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -279,14 +311,30 @@ impl Reading for P2P {
...
         match bincode::deserialize(message.as_ref()) {
-            Ok(tx) => self.recv_tx(tx).await,
+            Ok(protocol::Message::V0(msg)) => match msg {
+                protocol::MessageV0::Transaction(tx) => self.recv_tx(tx).await,
+                protocol::MessageV0::DiagnosticsRequest(kind) => {
+                    self.process_diagnostics_request(kind).await?
+                }
+                // Nodes are expected to ignore the diagnostics response.
+                protocol::MessageV0::DiagnosticsResponse(_) => (),
+            },
```
tuommaki, line 321: In the future, this should be extracted into separate function and later on possibly even into `protocol::Message::parse(bs)`.

crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -279,14 +311,30 @@ impl Reading for P2P {
...
 
+#[async_trait::async_trait]
+impl TxChannel for P2P {
+    async fn send_tx(&self, tx: &Transaction) -> Result<()> {
+        let msg = protocol::Message::V0(protocol::MessageV0::Transaction(tx.clone()));
+        let bs = Bytes::from(bincode::serialize(&msg)?);
+        self.broadcast(bs)?;
+        Ok(())
+    }
+}
```
tuommaki, line 337: Moved from the `mempool` and adapted for the new message type.

crates/node/src/networking/p2p/pea2pea.rs
musitdev: Yes, I think message protocol doesn't belong to the p2p domain. P2P is to exchange data between peers. It's more network and security related. Protocol data is to exchange data between node. It's more related to the node logic. 2 works, 2 separate domains.
For diagnostic, it's more about peer diag or node diag.

crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -44,12 +44,17 @@ pub struct P2P {
...
+    // Peer connection map: <(P2P TCP connection's peer address) , (peer's advertised address in peer_list)>.
+    // This mapping is needed for proper cleanup on OnDisconnect.
     peer_addr_mapping: Arc<tokio::sync::RwLock<HashMap<SocketAddr, SocketAddr>>>,
+    peer_list: Arc<tokio::sync::RwLock<BTreeSet<SocketAddr>>>,
+
     pub peer_http_port_list: Arc<tokio::sync::RwLock<HashMap<SocketAddr, Option<u16>>>>,
+
     http_port: Option<u16>,
+    nat_listen_addr: Option<SocketAddr>,
+    psk: Vec<u8>,
```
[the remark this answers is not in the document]
musitdev: Yes for the comment it's not very clear. You need to understand the algo to understand the message.

crates/node/src/networking/p2p/protocol.rs
musitdev: Yes. For the first implementation, I didn't know how far we can go in the conception and development.

PR #78, opened 2024-02-05
Ensure path exists under `/workspace`

tuommaki:
When downloading files to program VM, ensure that the configured destination path exists.


musitdev (approved):
I've done the fix in my PR to be able to test. I'll rebase and fix conflit.

PR #79, opened 2024-02-05
Fix ordering when persisting transaction from JSON-RPC

tuommaki:
Transaction persisting ordering is critical in order to guarantee all asset files to be present before transaction gets into execution.

- First transaction must be in DB so that foreign keys work correctly.
- Secondly transaction must be added into asset manager, which ensures that all necessary files are downloaded on node.
- Only as a last step it can be added into mempool, to be propagated via P2P network & later to be executed.

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -307,6 +307,7 @@ async fn run(config: Arc<Config>) -> Result<()> {
         database.clone(),
         mempool.clone(),
         asset_mgr.clone(),
+        database.clone(), // AclWhitelist impl.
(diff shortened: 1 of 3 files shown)


musitdev:
The 3 steps are not synchronous. The file download in step 2 is async in its own task. You can't force it to be done before Tx propagated.

musitdev (approved):
It decreases the risk. Decide if you merge it or close it.

tuommaki:
Going with this for now. We'll probably drop this soon anyway.

PR #82, opened 2024-02-08
Refactor task result files to streaming transfer

tuommaki:
Originally files resulting from program execution were embedded in `TaskResult` struct (with data & everything). With larger files this is problematic in many ways.

This commit refactors the result file transfer to take place one-by-one in streaming fashion. Now the `shim` will call `submit_file` for each file present in `TaskResult` and submit them before submitting the task result itself.

On server side, the file data is directly streamed into file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1200,6 +1200,7 @@ name = "gevulot-shim"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "async-stream",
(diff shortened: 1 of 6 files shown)


musitdev (approved):
In my current branch, I've made a little modification in this part to calculate the file checksum. I've done it on the VM side, when it will be merged I'll integrate it on the server side.

PR #83, opened 2024-02-12
Add program resource requirements to DB

tuommaki:
This change makes it possible to have `Program` specific resource requirements persisted in the DB.

### DIFF
--- a/crates/node/src/types/transaction.rs
+++ b/crates/node/src/types/transaction.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashSet, rc::Rc};
 
-use super::hash::Hash;
 use super::signature::Signature;
+use super::{hash::Hash, program::ResourceRequest};
 use eyre::Result;
 use libsecp256k1::{sign, verify, Message, PublicKey, SecretKey};
 use num_bigint::BigInt;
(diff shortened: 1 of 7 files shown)


crates/node/src/types/transaction.rs
```diff
@@ -127,6 +129,7 @@ impl Workflow {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
```
tuommaki, line 132:
```
133 | /   pub enum Payload {                                                                                                                                                       
134 | |       #[default]                                                                                                                                                           
135 | |       Empty,                                                                                                                                                               
136 | | /     Transfer {           
137 | | |         to: PublicKey,                                                         
138 | | |         value: BigInt,
139 | | |     },                   
    | | |_____- the second-largest variant contains at least 132 bytes
...   |        
146 | | /     Deploy {                                                                   
147 | | |         name: String,
148 | | |         prover: ProgramMetadata,
149 | | |         verifier: ProgramMetadata,
150 | | |     },                   
    | | |_____- the largest variant contains at least 344 bytes
...   |
170 | |       },
171 | |   }                          
    | |___^ the entire enum is at least 344 bytes 
```

I wonder if this is really an issue in any way? :thinking:

musitdev (approved):

PR #89, opened 2024-02-19
Add git metadata into gevulot node build

tuommaki:
Build git metadata into gevulot node binary for better diagnostics support.


tuommaki:
Going with this to get the git metadata into container. I want to build something on top of that. :wink:

PR #90, opened 2024-02-19
 Fixed the comment lines that overflowing from the screen.

hkey0:
Hello,
I have corrected a few comment lines that made it difficult to read the code because they did not fit on the screen.


tuommaki:
Hi! I appreciate your effort here, but I won't merge these at this time.

Reasons:
- These don't really improve the code base that much, but introduce unnecessary churn.
- This doesn't address one specific fix across whole code base but mixes couple different aspects.
- Regarding the comment style, they are actually on our todo list to make them aligned once and for all, but right now there are a lot more important issues we are working on.

PR #92, opened 2024-02-20
Rename node binary's key generation command

tuommaki:
`gevulot generate node-key` -> `gevulot generate key`

The `node-key` command still exists, but will be phased out. The new command name allows for more ubiquitous use of gevulot container e.g. in prover/verifier development purposes.

**NOTE:** I added bunch of you as reviewers just to make you aware of this change :slightly_smiling_face:


tuommaki:
Going with this at light speed. Shout me if you disagree with this and we'll revert.

kylegranger (approved):
LGTM :rocket:

PR #100, opened 2024-02-22
Refresh the deps

tuommaki:
We should do a proper cleanup + upgrade overhaul to all dependencies in bigger picture, but this change bumps up the deps at least within the specified limits. Small steps... Small steps.. :slightly_smiling_face:


tuommaki:
Build green, gotta overhaul the node's Cargo.toml so going with this.

PR #101, opened 2024-02-22
Restructure `Cargo.toml` for the node crate

tuommaki:
Node crate contains both, the node binary, but also the library for writing e.g. JSON-RPC clients or building transactions out of the node.

In order to avoid certain dependencies in the library (most notably the vsock related ones), those problematic dependencies are marked as optional and moved behind a feature flag that the binary build depends on.

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -4,52 +4,81 @@ version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 
-[[bin]]
-name = "gevulot"
-path = "src/main.rs"
-
 [dependencies]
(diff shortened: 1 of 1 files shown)


tuommaki:
Update: This change is needed in order to make `gevulot-cli` compile on Mac OS. Teemu just verified with this branch and it works. With `main` it fails on vsock API problems (operating system specific).

tuommaki:
Going with this so we can ensure the final blog post instructions.

tuommaki:
@musitdev Give me a shout if you see something fundamentally wrong here and we'll get it fixed then. :slightly_smiling_face:

PR #102, opened 2024-02-22
Only allocate GPU if requirements need it

tuommaki:
Now that it's possible to provide program specific resource requirements, the GPU should be only allocated when it's requested.

There's still plenty of room for improvement here (mainly allocating the right amount and fee GPUs), but this quick fix should help for the most immediate issues.

### DIFF
--- a/crates/node/src/vmm/qemu.rs
+++ b/crates/node/src/vmm/qemu.rs
@@ -244,7 +244,7 @@ impl Provider for Qemu {
             .args(["-qmp", &format!("tcp:localhost:{qmp_port},server")]);
 
         // TODO: When GPU argument handling is refactored, this should be fixed as well.
-        if self.config.gpu_devices.is_some() {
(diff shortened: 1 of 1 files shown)


crates/node/src/vmm/qemu.rs
```diff
@@ -244,7 +244,7 @@ impl Provider for Qemu {
             .args(["-qmp", &format!("tcp:localhost:{qmp_port},server")]);
 
         // TODO: When GPU argument handling is refactored, this should be fixed as well.
```
musitdev, line 246: The todo is still revelant?

musitdev (approved):

crates/node/src/vmm/qemu.rs
tuommaki: Yeah. It's a bit terse but the gist of that is basically that there's currently no tracking of _which_ GPU is allocated and which are available. This hack below shall work because the machines will only have one GPU for now, but in the future this needs improving once we get there.

PR #103, opened 2024-02-22
Correct vm output file path

musitdev:
This version correct the fact that the output files wasn't downloaded in the VM.
The path consist in sending to the VM the file relative path of the file in the node.
It's the same for input and output files.

For example, the generated proof path in the VM is: `/workspace/daa5279200428c6bcd9a586a686ba137ba2ea3434144c6e25538f3d1f06a62c5/5b96fa2c7915d790939bf74130edbe6b182f691e7dde996f3ac59a35a59642ea/proof.dat`

To correct the issue, I see 2 possibilities:
 * Change the grpc VM interface and define a local and node path for each file. Use the local path to copy in the VM and the node path to call `get_file()`.
 * In the `get_file()` call, get the VM associated task the same way as `get_task()` and map the local path to the node path to read the file in the node.
Tell me what you think.

Kyle can run his test with this PR to see it works using for the proof file in the VM using the path returns by `get_task_files_path()`.

During my correction, I see now that the Task are never stored in the DB. That why I remove the task table from the DB and all associated calls.

I've changed the Exec command input file parameter. I've defined a local_path and vm_path.

I've done some refactoring in the File.rs code to arrange File struct accordingly with the different domains(Tx, Exec Tx, Store).  

The Tx execution and file propagation work in my 3 nodes network.


crates/node/src/types/file.rs
```diff
@@ -8,54 +8,78 @@ use std::path::PathBuf;
...
     data_dir: &PathBuf,
-    task_id: &str,
     path: &str,
 ) -> Result<tokio::io::BufReader<tokio::fs::File>> {
     let mut path = Path::new(path);
     if path.is_absolute() {
         path = path.strip_prefix("/")?;
     }
-    let path = PathBuf::new().join(data_dir).join(task_id).join(path);
+    let path = PathBuf::new().join(data_dir).join(path);
```
tuommaki, line 17: Given I read the data flow correctly, here the path that VM returns is something that node is responsible for. It must not be this way. The VM must not have any idea of how the data is managed on the node side.

tuommaki (changes requested):
> To correct the issue, I see 2 possibilities:
>
> 1. Change the grpc VM interface and define a local and node path for each file. Use the local path to copy in the VM and the node path to call get_file().
> 2. In the get_file() call, get the VM associated task the same way as get_task() and map the local path to the node path to read the file in the node.
    Tell me what you think.

Definitely the second option. The end user nor the program running in the VM must **in no circumstances** have any need nor possibility to know about how the data is stored on the host. This would take all the freedom from the node storage implementation.

At design and API level the problem is very simple: The file paths that **end user**, **the transaction data** and **the program running in the VM** handle, are all from the VM filesystem namespace.

In other words: Any file path that the end user has any visibility or programmatic access, must be accessible as-is **inside the VM**. At the same time, the VM must not have any awareness of _anything_ that happens outside of the VM.

One extra note: Even though we provide the `shim` library that communicates over the gRPC protocol, I strongly believe that there will be times when the deployed program implements the gRPC client directly by itself for whatever reason and because of this, the protocol must be kept simple and abstract. It must not leak any node-internal information into the VM.

The way I see this, the file management should be rather simple:
- As said above, all file paths in transactions and in VM are VM-internal. Practically they are always `/workspace/<something>`, because that's the only writable location inside the VM.
- On host side, all files are transaction specific. Whenever files are handled, there must be transaction hash known at that point.
  - Physically for now, the files are on the host file system. On some path of `${transaction_specific_directory}/${vm_understandable_file_path}` that `${transaction_specific_directory}` used to be `Path::new(data_dir).join(tx_hash).join(vm_understandable_file_path)` _I think?_ Ideally, those would be grouped under something like `$data_dir/txdata/<tx hash>/...`.

tuommaki (approved):
LGTM :+1:

crates/cli/src/lib.rs
```diff
@@ -48,9 +48,11 @@ struct JsonExecArgs {
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub enum JsonProgramData {
     Input {
-        //path to the local file or Hash os the distant file
-        file: String,
-        //optional url to the file if the file hash is provided by file
+        // path to the local file or Hash os the distant file
+        local_path: String,
+        // Path of the file in the VM containing /warspace directory
```
tuommaki, line 53:
```suggestion
        // Path of the file in the VM containing /workspace directory
```

crates/node/src/types/file.rs
```diff
@@ -6,114 +6,201 @@ use serde::Serialize;
...
-    if path.is_absolute() {
-        path = path.strip_prefix("/")?;
-    }
-    let path = PathBuf::new().join(data_dir).join(task_id).join(path);
-    let fd = tokio::fs::File::open(path).await?;
-    Ok(tokio::io::BufReader::new(fd))
+// Describe a file use by an executed task.
+#[derive(Clone, Debug)]
+pub struct TaskVmFile<E> {
+    //    pub node_file_path: String,
```
tuommaki, line 12:
```suggestion
```

crates/node/src/vmm/vm_server.rs
```diff
@@ -157,7 +185,8 @@ impl VmService for VMServer {
                                 }))
                                 .await
                             {
-                                tracing::error!("send {} bytes from file {}: {}", n, &req.path, &e);
+                                //                                tracing::error!("send {} bytes from file {}: {}", n, &req.path, &e);
```
tuommaki, line 188:
```suggestion
```

tuommaki (approved):
LGTM :+1: 

Couple very minor nitpicks (mostly commented-out code).

PR #104, opened 2024-02-24
Fix container build

tuommaki:
Due to separation of dependencies between Gevulot node binary and library, now the required feature flag (`node-binary`) must be specified in order to yield the node binary from the build.


tuommaki:
Build green, going with this to yield a fresh container :slightly_smiling_face:

PR #105, opened 2024-02-26
Consolidate `Scheduler` state mutexes in one

tuommaki:
Multiple distinct mutexes on related data structures introduce high risk to run into dead lock (which has happened in some tests).

This change consolidates four related collections into one `SchedulerState` to put them under one lock to avoid inter-dependent conflicts.


tuommaki:
Superseded by #107

PR #110, opened 2024-02-27
Improve diagnostics message in P2P

tuommaki:
- In response message, some of the wrapping elements were missing.
- Added public node key to response, so that each node / response can be identified.

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -180,6 +180,8 @@ impl workflow::TransactionStore for storage::Database {
 }
 
 async fn run(config: Arc<Config>) -> Result<()> {
+    let node_key = read_node_key(&config.node_key_file)?;
+
     let database = Arc::new(Database::new(&config.db_url).await?);
 
     let http_peer_list: Arc<tokio::sync::RwLock<HashMap<SocketAddr, Option<u16>>>> =
@@ -213,11 +215,13 @@ async fn run(config: Arc<Config>) -> Result<()> {
         });
     }
 
+    let public_node_key = PublicKey::from_secret_key(&node_key);
     let p2p = Arc::new(
         networking::P2P::new(
             "gevulot-p2p-network",
             config.p2p_listen_addr,
             &config.p2p_psk_passphrase,
+            public_node_key,
             Some(config.http_download_port),
             config.p2p_advertised_listen_addr,
             http_peer_list,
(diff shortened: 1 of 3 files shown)


crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -163,16 +167,19 @@ impl P2P {
         source: SocketAddr,
         req: protocol::DiagnosticsRequestKind,
     ) -> io::Result<()> {
-        let resp = protocol::DiagnosticsResponseV0::Version {
```
tuommaki, line 166: This was a bit bare bone :grimacing:

crates/node/src/main.rs
```diff
@@ -213,11 +215,13 @@ async fn run(config: Arc<Config>) -> Result<()> {
         });
     }
 
+    let public_node_key = PublicKey::from_secret_key(&node_key);
```
musitdev, line 218: A remark about how private keys are stored in the memory. As the run() function never returns, the `node_key` isn't drop and always stay in the node memory.  For Devnet, it's not an issue, but more generally it's something we've to take care. Usually, private key are zeroed after being used.

crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -163,16 +167,19 @@ impl P2P {
...
-            build: format!(
-                "{}: {}",
-                env!("VERGEN_BUILD_TIMESTAMP"),
-                env!("VERGEN_GIT_DESCRIBE")
-            ),
-        };
+        let resp = protocol::Message::V0(protocol::MessageV0::DiagnosticsResponse(
+            self.public_node_key,
+            protocol::DiagnosticsResponseV0::Version {
+                major: env!("CARGO_PKG_VERSION_MAJOR").parse::<u16>().unwrap(),
```
musitdev, line 173: These variable are only set is you use cargo. Node are not always started with cargo run?
The other possibility is to use the last commit that can be set in the code during the build process.

musitdev (approved):
Some general remarks.

crates/node/src/networking/p2p/pea2pea.rs
tuommaki: These are build into the binary:
- https://github.com/gevulotnetwork/gevulot/blob/main/crates/node/build.rs#L2-L5
- https://github.com/gevulotnetwork/gevulot/blob/main/Containerfile#L6-L7

PR #111, opened 2024-02-28
Add flag for `--no-execution` to disable tx exec

tuommaki:
This change allows running a full node that doesn't execute transactions.

Transaction execution is disabled by decoupling event-based transaction handling from the execution backend by introducing an intermediary trait in the `txvalidation` module.

When `--no-execution` flag is specified, nothing from execution backend is started (no scheduler, no provider and no support functionality related to them). Instead, a database backend implementation for that intermediary trait is implemented and once the transaction is fully validated and assets downloaded, it's only saved in the databased and then dismissed.

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -45,6 +45,14 @@ pub struct Config {
 
     #[arg(
         long,
+        long_help = "No execution flag. When set as true, the node does not execute transactions.",
+        env = "GEVULOT_NODE_NO_EXECUTION",
+        default_value_t = false
+    )]
+    pub no_execution: bool,
+
+    #[arg(
+        long,
         long_help = "File where the node key is persisted",
         env = "GEVULOT_NODE_KEY_FILE",
         default_value_os_t = PathBuf::from("/var/lib/gevulot/node.key"),
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -8,6 +8,7 @@ use cli::{
 };
 use eyre::Result;
 use gevulot_node::types;
+use gevulot_node::types::transaction::Received;
 use libsecp256k1::{PublicKey, SecretKey};
 use pea2pea::Pea2Pea;
 use rand::{rngs::StdRng, SeedableRng};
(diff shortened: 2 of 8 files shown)


tuommaki:
This is a draft for now. I'll test this a little bit later once I have built basic terraform setup for this. Ready for review however.

musitdev (commented):
I think you can do it differently. Tx vlaidation, Tx exec and Storage are 3 separate domains. You need only the Tx validation and Storage. So the other possibility is to remove the Tx execution.
To do that, you use directly the: 
`txvalidation::TxEventSender::<txvalidation::TxResultSender>::build(tx_sender.clone())`
in an async loop to get the Tx from it and store them in the database.
This way you skip the execution, you don't start it and more important you don't have to modify it because of a more general need.
Execution layer don't have to know that want to execute a full archive node.

In this model, we skip the mempool because it's too linked to the Execution layer and do nothing, just keep Tx until the scheduler get them.
The better way is to define a mempool that get validated Tx, propagate them if needed (tx propagation is part of the mempool domain) and make them available to the other domain: Tx execution / Storage / ... using stream.

This way you can activate Tx execution, Storage or not depending on the global needs. Each layer don't have to know how they are used.

tuommaki:
>I think you can do it differently. Tx vlaidation, Tx exec and Storage are 3 separate domains. You need only the Tx validation and Storage. So the other possibility is to remove the Tx execution.
>To do that, you use directly the:
`txvalidation::TxEventSender::<txvalidation::TxResultSender>::build(tx_sender.clone())`
in an async loop to get the Tx from it and store them in the database.

Where would I use that?

> This way you skip the execution, you don't start it and more important you don't have to modify it because of a more general need.
> Execution layer don't have to know that want to execute a full archive node.
>
> In this model, we skip the mempool because it's too linked to the Execution layer and do nothing, just keep Tx until the scheduler get them.

If mempool is skipped, how does the transaction gets persisted?

> The better way is to define a mempool that get validated Tx, propagate them if needed (tx propagation is part of the mempool domain) and make them available to the other domain: Tx execution / Storage / ... using stream.
>
> This way you can activate Tx execution, Storage or not depending on the global needs. Each layer don't have to know how they are used.

I don't fully understand your suggestion at concrete level, but it sounds like you'd have a simpler / better solution than I have. Can you elaborate a bit so I can get on the same page :slightly_smiling_face:

tuommaki:
This second revision should be now pretty much the same change that you suggested in your PR.

musitdev (approved):

PR #115, opened 2024-02-29
Log P2P connect error

tuommaki:
Ignoring the `node.connect()` error hides a problem if e.g. the peer doesn't allow connection in to advertised P2P address.


tuommaki:
Going .. to get this in container. Give me a shout if there's something wrong etc. :pray:

PR #118, opened 2024-03-01
Improve txfile URL handling

tuommaki:
The address generated for txfile URL originates from the P2P bind address. In case the node is behind NAT, the address should take this into account.

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -258,7 +258,7 @@ async fn run(config: Arc<Config>) -> Result<()> {
         "http://{}:{}",
         config
             .p2p_advertised_listen_addr
-            .unwrap_or_else(|| p2p_listen_addr)
+            .unwrap_or(p2p_listen_addr)
             .ip(),
         config.http_download_port
     );


crates/node/src/main.rs
```diff
@@ -253,7 +256,10 @@ async fn run(config: Arc<Config>) -> Result<()> {
     let workflow_engine = Arc::new(WorkflowEngine::new(database.clone()));
     let download_url_prefix = format!(
         "http://{}:{}",
-        config.p2p_listen_addr.ip(),
+        config
+            .p2p_advertised_listen_addr
+            .unwrap_or_else(|| p2p_listen_addr)
+            .ip(),
```
tuommaki, line 262: This accounts for the NAT scenario.

There is still a problem when node is configured to be bound to all interfaces (`0.0.0.0`), because for the peer that address doesn't tell anything. I'm very hesitant on relying on any of public services for finding out the node's public IP so I'm wondering if we could improve the P2P handshake in a way that the peer would report back the public IP and that would then find its way to places where the URLs are constructed?

crates/node/src/main.rs
```diff
@@ -253,7 +256,10 @@ async fn run(config: Arc<Config>) -> Result<()> {
     let workflow_engine = Arc::new(WorkflowEngine::new(database.clone()));
     let download_url_prefix = format!(
         "http://{}:{}",
-        config.p2p_listen_addr.ip(),
+        config
+            .p2p_advertised_listen_addr
+            .unwrap_or(p2p_listen_addr)
+            .ip(),
```
tuommaki, line 262: This accounts for the NAT scenario.

There is still a problem when node is configured to be bound to all interfaces (0.0.0.0), because for the peer that address doesn't tell anything. I'm very hesitant on relying on any of public services for finding out the node's public IP so I'm wondering if we could improve the P2P handshake in a way that the peer would report back the public IP and that would then find its way to places where the URLs are constructed?

musitdev: Here we need a service that gives network information about the node. It can be hosted by the p2p because p2p need it, or it can be a more independent service.
The role of this service is to give the node host data (IP address for example) to other module depending on the network use:
 * peer connection
 * Tx internode download
 * RPC data download.
 ...
 This service has the logic that define the best IP to use depending on the needs. For example, if we get an RPC request from a local network, the RPC URL should be the local bind address and not the public. Public address are not always routed locally.
 There can be a lot of cases, so we can put everything in a module that grow little by little depending on the new case we see.
 The API is: get best node host IP depending on a need, and it returns the best IP to use depending on the context.
 We can start with simple service that return the public IP and define new rule little by little.
 For the specific case you have, the rule can be if p2p_listen_addr is 0.0.0.0 you can use the caller IP to get the download address: if local return local interface IP if not public IP. For p2p node connect, we can define something different or the same.

tuommaki: Yup. Something like that. I'll go with this fix though, as that alleviates the most urgent need for devnet. But we should definitely work on this network service next week, I think. In MVP style, feature by feature. Not building everything at once :slightly_smiling_face:

PR #122, opened 2024-03-05
Fetch program with resource requests if exists

tuommaki:
`Database.find_program(..)` is used e.g. in program manager, which depends on embedded program resource requests, if such are configured.

Old version of this method didn't fetch them along the program data, so the method is now a wrapper for `get_program(..)` but handles error as non-existing program.


tuommaki:
Successfully tested. Gotta move on to get the fix for the node operators.

PR #123, opened 2024-03-05
Merge ACL whitelist

musitdev:
I let the code in the network module. Not sure where to put it.
I define the merge function alone so that I can test it without the db and network.
I put some rules about the merge:
 * if a pubkey fail to parse skip it and continue the merge without it.
 * if the network fetches fail for one line, abort and do nothing.
 * if the new acl key list is empty, do nothing instead of removing all the keys.

### DIFF
--- a/crates/node/src/networking/mod.rs
+++ b/crates/node/src/networking/mod.rs
@@ -112,15 +112,14 @@ where
         }
     }
 
-
     tracing::info!("{key_count} new keys whitelisted and {removed_key_count} keys removed",);
     Ok(())
 }
 
 #[cfg(test)]
 mod tests {
-    
-use super::*;
+
+    use super::*;
     use tokio::sync::Mutex;
 
     struct TestDb(Mutex<HashSet<String>>);
@@ -137,12 +136,10 @@ use super::*;
                 .map_err(|err| err.into())
         }
         async fn acl_whitelist(&self, key: crate::entity::PublicKey) -> Result<()> {
-            println!("insert:{}", key.to_string());
             self.0.lock().await.insert(key.to_string());
             Ok(())
         }
         async fn acl_deny(&self, key: &crate::entity::PublicKey) -> Result<()> {
-            println!("remove:{}", key.to_string());
             let str_key = key.to_string();
             self.0.lock().await.remove(&str_key);
             Ok(())
@@ -196,7 +193,7 @@ use super::*;
             assert!(set.get("040f28123df7a638647d867dfe186395999a276048c76c1bd56d4b792eca56449751944d6cd0b208f44b9aec332b4b6d0630406ebb07ccfb17e7c505a07156a792").is_some());
         }
     }
-    
+
     #[tokio::test]
     async fn test_acl_merge_with_empty() {
         let db = TestDb(Mutex::new(HashSet::new()));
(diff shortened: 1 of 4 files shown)


crates/node/src/networking/mod.rs
```diff
@@ -51,3 +52,189 @@ impl WhitelistSyncer {
...
+    S: Stream<Item = Result<B, std::io::Error>> + Unpin,
+    B: Buf,
+{
+    let mut lines = reader.lines();
+    let db_keys = database.get_acl_whitelist().await?;
+    let mut db_keys: HashSet<crate::entity::PublicKey> = HashSet::from_iter(db_keys.into_iter());
+
+    let mut new_keys = vec![];
+    // Received list shouldn't be empty.
+    let mut non_empry_list = false;
```
tuommaki, line 83:
```suggestion
    let mut non_empty_list = false;
```

crates/node/src/networking/mod.rs
```diff
@@ -51,3 +52,189 @@ impl WhitelistSyncer {
...
+    // If an error occurs during fetch cancel merge to avoid removing all db keys.
+    while let Some(line) = lines.next_line().await? {
+        non_empry_list = true;
```
tuommaki, line 86:
```suggestion
        non_empty_list = true;
```

crates/node/src/networking/mod.rs
```diff
@@ -51,3 +52,189 @@ impl WhitelistSyncer {
...
+    }
+
+    let key_count = new_keys.len();
+    for public_key in new_keys {
+        database.acl_whitelist(public_key).await?;
+    }
+    let removed_key_count = db_keys.len();
+
+    // Do not remove all list if received list is empty.
+    if non_empry_list {
```
tuommaki, line 109:
```suggestion
    if non_empty_list {
```

crates/node/src/storage/database/postgres.rs
```diff
@@ -535,6 +535,14 @@ impl Database {
         Ok(())
     }
 
+    pub async fn get_acl_whitelist(&self) -> Result<Vec<entity::PublicKey>> {
+        let dbkey_set: Vec<entity::PublicKey> = sqlx::query("SELECT 1 FROM acl_whitelist")
```
tuommaki, line 539: Did you test this? This query returns literal `1` for each row that exists in the DB. I'd expect this to yield error when trying to parse as public key?

tuommaki (commented):
In general the code looks good, but I think the DB query requires changes.

crates/node/src/storage/database/postgres.rs
musitdev: Yes, I copy and paste from another query. I thought the 1 was interpreted as a prepared statement, the first return field. I didn't test the read in fact because I start only one time.
I correct and to a read test.

musitdev: I've to add a test too, when the public key parse fail.

musitdev: I've corrected the query and added the test.

tuommaki (approved):
LGTM :+1:

Please apply the typo fix for the variable name :wink:

musitdev:
Done I merge

PR #124, opened 2024-03-06
Set vmid txhash link

musitdev:
Most, I change to use Tx Hash to link task/Tx and Vm execution.
I remove VMId when it wasn't necessary for the logic.
I didn't touch the resource part.

I do a local test and the Txs execute correctly.


tuommaki (approved):
LGTM :+1:

I also tested this with e2e test & taiko-demo. Both succeeded :ok_hand:

PR #132, opened 2024-03-07
Report error from `submit_file()` in shim

tuommaki:
If an error occurs during file transfer from VM to node, an error should be reported from the function to higher level so that the program and VM can terminate deterministically.


musitdev (commented):
I've done the same change but differently:
https://github.com/gevulotnetwork/gevulot/compare/main...transfert_big_files
I think both work, but in mine there's no Mutex.

tuommaki:
> I've done the same change during my tests but differently:
> https://github.com/gevulotnetwork/gevulot/compare/main...transfert_big_files
> I think both work, but in mine there's no Mutex.

I think your solution is a lot better. I switched to that.

tuommaki:
I recon this is ok to merge since it works and the original code is from you @musitdev ?

PR #136, opened 2024-03-13
Change Vsock to VirFs to tranfer files between the VM and the node.

musitdev:
Remove Vsock declarations in the Qemu VM start args.
Add VirtFs arguments.
Update file transfer logic.
Change the result management between VM and node to report more VM error in the submit result.

### DIFF
--- a/crates/node/src/types/file.rs
+++ b/crates/node/src/types/file.rs
@@ -18,6 +18,14 @@ impl<E> TaskVmFile<E> {
         &self.vm_file_path
     }
 }
+impl TaskVmFile<()> {
+    pub fn get_workspace_path(data_directory: &Path, tx_hash: Hash) -> PathBuf {
+        PathBuf::new()
+            .join(data_directory)
+            .join(&tx_hash.to_string())
+            .join(gevulot_shim::WORKSPACE_NAME)
+    }
+}
 
 // Define A task file send to the VM. Extension contains the node file path.
 #[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq)]
--- a/crates/node/src/types/file.rs
+++ b/crates/node/src/types/file.rs
@@ -22,7 +22,7 @@ impl TaskVmFile<()> {
     pub fn get_workspace_path(data_directory: &Path, tx_hash: Hash) -> PathBuf {
         PathBuf::new()
             .join(data_directory)
-            .join(&tx_hash.to_string())
+            .join(tx_hash.to_string())
             .join(gevulot_shim::WORKSPACE_NAME)
     }
 }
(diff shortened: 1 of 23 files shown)


crates/tests/test-programs/src/prover.rs
```diff
@@ -7,27 +8,30 @@ fn main() -> Result<()> {
...
 
-    std::fs::write("/workspace/proof.dat", b"this is a proof.")?;
-
+    std::fs::write(format!("{WORKSPACE_PATH}/proof.dat"), b"this is a proof.")?;
     task.result(
         vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
-        vec![String::from("/workspace/proof.dat")],
+        vec![
+            String::from(format!("{WORKSPACE_PATH}/proof.dat")),
+            String::from(format!("{WORKSPACE_PATH}/bigfile.bin")),
```
tuommaki, line 34: Should this file be generated during runtime?

tuommaki (commented):
LGTM in general. Couple very minor nits / fix suggestions and one question about the tests prover's bigfile.

crates/node/src/types/file.rs
```diff
@@ -18,6 +18,14 @@ impl<E> TaskVmFile<E> {
         &self.vm_file_path
     }
 }
+impl TaskVmFile<()> {
+    pub fn get_workspace_path(data_directory: &Path, tx_hash: Hash) -> PathBuf {
+        PathBuf::new()
+            .join(data_directory)
+            .join(&tx_hash.to_string())
```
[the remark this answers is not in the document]
musitdev: Yes I agree with you. I add the same idea but I didn't change because I wanted to keep things like they were.
I can add a directory in this PR or open a new PR to do the change?
I prefer data or files.

crates/tests/test-programs/src/prover.rs
musitdev: I wanted an easy way to transfer a big file from prover to verifier for test, so I was thinking of using the same file was easier.
It was only for test, and at some point perhaps we should revert to a more playful example.

crates/node/src/types/file.rs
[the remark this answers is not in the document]
tuommaki: >  open a new PR to do the change?

Open a new PR so it's clearer that way :+1:

> I prefer data or files.

Both work for me. Possibly prefix with "tx" so it's absolutely obvious what it's about. :+1:

crates/tests/test-programs/src/prover.rs
tuommaki: Yeah, I get that. I was just thinking if this code works as-is on a new installation? I mean, would the end user need to put that "bigfile.bin" somewhere in order to make this work or does that somehow work already out-of-the-box?

tuommaki (approved):
If everything works, let's roll with this :slightly_smiling_face:

PR #137, opened 2024-03-13
Change Tx validation logic to add: Tx wait that its parent has been validated.

musitdev:
A Child Tx must be executed after its parent if it has a parent Tx.
This order is not guaranteed by the current code and can easily arrive when Proof has a big file as input. The child verify Tx can be executed during the proof input file download.
After Tx verification, the Tx's parent, if it exists, is verified to be already validated.
To do that, the presence of the parent Tx in the db is verified. A cache is used to avoid Db query of the recent Tx.
If the parent is not found, the Tx is put in a waiting list.
When the parent arrives, the child Tx are removed from the waiting list and added to the mempool with the parent tx.

The process_event logic code is more complex because I wanted to avoid using a mutex for the cache, and we need to mix sync code (cache / waiting Tx verification) and async one.

I set it in draft because to be fully tested, the PE #136 must be merged to be able to test it with bigfile.

### DIFF
--- a/crates/node/src/txvalidation/event.rs
+++ b/crates/node/src/txvalidation/event.rs
@@ -222,7 +222,8 @@ impl TxEvent<WaitTx> {
         let new_txs = new_tx
             .map(|tx| {
                 let mut ret = cache.remove_waiting_children_txs(&tx.tx.hash);
-                ret.push(tx);
+                // Add the parent Tx first to be processed the first.
+                ret.insert(0, tx);
                 ret
             })
             .unwrap_or(vec![]);
--- a/crates/node/src/txvalidation/event.rs
+++ b/crates/node/src/txvalidation/event.rs
@@ -197,8 +197,8 @@ impl TxEvent<WaitTx> {
         cache: &mut TXCache,
         storage: &impl Storage,
     ) -> Result<Vec<TxEvent<NewTx>>, EventProcessError> {
-        // Verify Tx'x parent is already present.
-        let new_txs = if let Some(parent) = self.tx.payload.get_parent_tx() {
+        // Verify Tx'x parent is present or not.
+        let new_tx = if let Some(parent) = self.tx.payload.get_parent_tx() {
             if cache.is_tx_cached(parent)
                 || storage
                     .get(parent)
@@ -206,19 +206,28 @@ impl TxEvent<WaitTx> {
                     .map_err(|err| EventProcessError::StorageError(format!("{err}")))?
                     .is_some()
             {
-                let mut new_txs = cache.remove_waiting_children_txs(parent);
-                new_txs.push(self);
-                new_txs
+                // Present return the Tx
+                Some(self)
             } else {
                 //parent is missing add to waiting list
                 cache.add_new_waiting_tx(*parent, self);
-                vec![]
+                None
             }
         } else {
             //no parent always new Tx.
-            vec![self]
+            Some(self)
         };
 
+        //remove child Tx if any from waiting list.
+        let new_txs = new_tx
+            .map(|tx| {
+                let mut ret = cache.remove_waiting_children_txs(&tx.tx.hash);
+                ret.push(tx);
+                ret
+            })
+            .unwrap_or(vec![]);
+
+        // Add new tx to the cache.
         let ret = new_txs
             .into_iter()
             .map(|tx| {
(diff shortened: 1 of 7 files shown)


crates/node/src/txvalidation/event.rs
```diff
@@ -202,3 +273,275 @@ impl TxEvent<NewTx> {
             .await
     }
 }
+
+// Use to cache New tx and store waiting Tx that have missing parent.
+pub struct TXCache {
```
tuommaki, line 278: Should this possibly be named `TxCache` so that it's aligned with other types (e.g. `TxEvent` or `WaitTx` etc.)?

crates/node/src/txvalidation/event.rs
```diff
@@ -202,3 +273,275 @@ impl TxEvent<NewTx> {
...
+impl TXCache {
+    pub fn new() -> Self {
+        Self::build(
+            MAX_CACHED_TX_FOR_VERIFICATION,
+            MAX_WAITING_TX_FOR_VERIFICATION,
+            MAX_WAITING_TIME_IN_MS,
+        )
+    }
+    pub fn build(cache_size: usize, max_waiting_tx: usize, max_waiting_time: u64) -> Self {
+        let cachedtx_for_verification =
```
tuommaki, line 298:
```suggestion
        let cached_tx_for_verification =
```
Would it be ok to separate the words like this so it's easier to read when skimming the code through?

crates/node/src/txvalidation/event.rs
```diff
@@ -202,3 +273,275 @@ impl TxEvent<NewTx> {
...
+        assert_eq!(wait_tx_cache.waiting_tx.len(), 2);
+
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        // Add new Tx to evict the old one.
+        let tx_event = new_proof_tx_event(parent1_hash);
+        let tx_hash = tx_event.tx.hash;
+        let res = tx_event.process_event(&mut wait_tx_cache, &db).await;
+        assert!(res.is_ok());
+        // Evicted but new ket added.
```
tuommaki, line 470: What's `ket` here?

crates/node/src/txvalidation/event.rs
```diff
@@ -202,3 +273,275 @@ impl TxEvent<NewTx> {
...
+        assert!(res.is_ok());
+        let ret_txs = res.unwrap();
+        // Return only the parent Tx. The other has been evicted during parent Tx add.
+        assert_eq!(ret_txs.len(), 2);
+        assert!(ret_txs[0].tx.hash == parent1_hash || ret_txs[1].tx.hash == parent1_hash);
+        assert!(ret_txs[0].tx.hash == tx_hash || ret_txs[1].tx.hash == tx_hash);
+    }
+
+    #[tokio::test]
+    async fn test_waittx_process_event() {
```
tuommaki, line 486:
```suggestion
    async fn test_wait_tx_process_event() {
```

crates/node/src/txvalidation/mod.rs
```diff
@@ -146,52 +155,96 @@ pub async fn spawn_event_loop(
...
+                    }
+                    // Verify Tx parent and send to mempool all ready Tx.
+                    Some(Ok((wait_tx_res, callback))) = validated_txs_futures.next() =>  {
+                        match wait_tx_res {
+                            Ok(wait_tx) => {
+                               match wait_tx.process_event(&mut wait_tx_cache, storage.as_ref()).await {
+                                    Ok(new_tx_list) => {
+                                        //
+                                        let jh  = tokio::spawn({
+                                            let newtx_receiver = newtx_receiver.clone();
```
tuommaki, line 209: Maybe these could also be renamed to `new_tx_receiver` for easier readability?

crates/node/src/types/transaction.rs
```diff
@@ -193,6 +193,15 @@ impl std::fmt::Display for Payload {
 }
 
 impl Payload {
+    //return the parent tx associated to the payloas if any.
```
tuommaki, line 196:
```suggestion
    // Return the parent tx associated to the payloads, if any.
```

tuommaki (commented):
LGTM in general. I didn't fully execute the code in my head, but skimmed it through. Couple minor suggestions on naming mostly.

When you are done with this, can you change the state to `Ready for Review` and  I'll do one more look on changed parts :slightly_smiling_face:

crates/node/src/txvalidation/event.rs
musitdev: I've updated all occurrences in the code.

musitdev:
I've applied the changes.

tuommaki (approved):
Awesome :raised_hands:

PR #139, opened 2024-03-14
Handle P2P discovery connection error

tuommaki:
If the initial discovery nodes cannot be connected, handle the error correctly. If no discovery nodes respond, then terminate the program.

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -292,7 +292,7 @@ async fn run(config: Arc<Config>) -> Result<()> {
 
     // If we couldn't connect to any P2P nodes, we'll be left out alone forever.
     // Useless to run at that point.
-    if connected_nodes == 0 {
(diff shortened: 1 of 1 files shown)


crates/node/src/main.rs
```diff
@@ -281,6 +290,13 @@ async fn run(config: Arc<Config>) -> Result<()> {
         }
     }
 
+    // If we couldn't connect to any P2P nodes, we'll be left out alone forever.
+    // Useless to run at that point.
+    if !config.p2p_discovery_addrs.is_empty() && connected_nodes == 0 {
```
tuommaki, line 295: For the P2P bootstrap node, the discovery addresses are empty as that's the beacon of the network. In that case it's fine that `connected_nodes == 0`. For the others: It's useless to continue when no friends are playing with. :slightly_smiling_face:

musitdev (approved):

PR #141, opened 2024-03-15
Drop `bigfile.bin` from dummy test prover's result

tuommaki:
The `bigfile.bin` doesn't exist nor is it created during the test and therefore it breaks the E2E test. Test works after dropping it.


tuommaki:
These E2E tests must not break as they are referenced in our blog post as well: https://blog.gevulot.com/i/141891862/test-proof-generation-and-verification-with-our-simple-end-to-end-test

musitdev:
Yes I'm going to develop my own prover/verifier test in a separate project.

musitdev (approved):

PR #142, opened 2024-03-15
manage new peer connection in a specific task

musitdev:
To avoid cancelling the handshake if an error occurs during a new declared peer connection, these new connections are executed in a specific task.


tuommaki (approved):
LGTM :+1:

PR #143, opened 2024-03-18
Wait for QEMU to prevent zombies

tuommaki:


musitdev (approved):
Perhaps you can do the kill and wait on the same option and avoid the unwrap() and expect() like:
```
            qemu_vm_handle
                .child
                .as_mut()
                .ok_or(std::io::Error::other(
                    "No child process defined for this handle",
                ))
                .and_then(|p| {
                    p.kill()?;
                    p.wait()
                })?;
```

tuommaki:
> Perhaps you can do the kill and wait on the same option and avoid the unwrap() and expect()

Yeah, that was a good suggestion :+1:

I changed the code to your version.

PR #144, opened 2024-03-18
Correct file missing bug. Add file copy to virtfs before executing VM

musitdev:
The task's files weren't copied to the virtfs location before the VM execution.
I add this copy before the VM start to be sure the files are copied when the VM is started.
If the copy fails, the task and associated Tx are cancelled and marked as executed.

I think we should change how files are stored because there are several places depending on the file's Tx, and it creates useless complexity.
For example:
 * deploy tx: image file under /images/<file checksum>/<image name>.
 * proof Tx: input file under /<tx Hash>/<VM file path>: this path works directly when the Tx is executed because it's the Virtfs path of the file.
 * Tx output file are moved from Virtfs location: /<tx Hash>/workspace/<file VM path> to  /<new tx hash>/<file checksum>/<file_name>.
 * verify tx: input file under /<tx hash>/<file checksum>/<file_name>: these files need to be copied to virtfs localtion when the Tx is executed. It's the current PR update.

I think we should use:
 * for image: /images/<file checksum> I'll remove the image name to avoid name incompatibility with file system path.
 * /txfiles/<tx hash>/<file checksum>. The same, I'll remove the filename or path to avoid incompatibilities. These incompatibilities can be a vector of attack.
 * Virtfs path: /vmworkspace/<tx hash>/workspace
Filename in the path can ease the development. Perhaps we can define a dev mode where the filenames are added?

Currently, with the current file path definition, there's an issue that can arrive in a very particular case:
 * The input file of the proof Tx is also the output file of the tx.
 * This file is moved to the verifyTx path, so the proof Tx lose its input file.
 * Another node request the proof Tx input that is missing. The Tx fail on the other node.

Last I remove the <firtfs>/worplace cleanup because it can remove the proof fs input file.

So with this correction if works, but I think we should change the way file are managed. I can open an issue about it. There's already an issue about txfile root path for tx files that I've worked on.

### DIFF
--- a/crates/node/src/scheduler/mod.rs
+++ b/crates/node/src/scheduler/mod.rs
@@ -240,6 +240,27 @@ impl Scheduler {
                 }
             };
 
+            //copy Task file to VM workspace
+            //Validate that everything is ok before starting the task.
+            for file in task.files.iter() {
+                if let Err(err) = file
+                    .copy_file_for_vm_exe(&self.data_directory, task.tx)
+                    .await
+                {
+                    tracing::error!(
+                        "An error occurs during Tx:{} task file copy for VM execution {err}",
+                        task.tx
+                    );
+                    if let Err(err) = self.database.mark_tx_executed(&task.tx).await {
+                        tracing::error!(
+                            "failed to update transaction.executed => true - tx.hash: {}",
+                            task.tx
+                        );
+                    }
+                    continue;
+                }
+            }
+
             // Push the task into program's work queue.
             {
                 let mut state = self.state.lock().await;
--- a/crates/node/src/scheduler/mod.rs
+++ b/crates/node/src/scheduler/mod.rs
@@ -240,8 +240,8 @@ impl Scheduler {
                 }
             };
 
-            //copy Task file to VM workspace
-            //Validate that everything is ok before starting the task.
+            // Copy Task file to VM workspace
+            // Validate that everything is ok before starting the task.
             for file in task.files.iter() {
                 if let Err(err) = file
                     .copy_file_for_vm_exe(&self.data_directory, task.tx)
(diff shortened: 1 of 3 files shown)


crates/node/src/scheduler/mod.rs
```diff
@@ -240,6 +240,27 @@ impl Scheduler {
...
+                    tracing::error!(
+                        "An error occurs during Tx:{} task file copy for VM execution {err}",
+                        task.tx
+                    );
+                    if let Err(err) = self.database.mark_tx_executed(&task.tx).await {
+                        tracing::error!(
+                            "failed to update transaction.executed => true - tx.hash: {}",
+                            task.tx
+                        );
+                    }
```
tuommaki, line 259: I'm a bit unsure about this. If task execution in VM fails, we don't mark the tx as executed and that has pros and cons. So far my thinking has been that it's good as it can be retried (after the technical error has been fixed).

This one is similar: One obvious case is when disk is full. The transaction could be executed when the disk has been cleaned.

On the other hand there can be a corner case where transaction could get stuck forever, which is not ideal either, but these individual exceptions can be manually fixed from DB.

WDYT?

tuommaki (approved):
LGTM; One question on the tx executed flag use though.

crates/node/src/scheduler/mod.rs
musitdev: I agree, Tx re-execution is not easy to define because it depends on the cause. In our implementation the logic is simple, the node tries to re-execute a little later.  It leads to Tx that re-execute all the time until the reason is solved or the node restarted.
I think we can consider that if the Tx fail on one node, there's 2 case:
 * it fails only on that node (disk full for example), the user can get Tx result on other nodes. The node should be upgraded and do a bootstrap. Hot upgrade without a correct bootstrap can lead to inconsistency between nodes.
 * it fails on all node and the user has to change the Tx and retry. It's a normal case that don't lead to inconsistency between nodes.
As we didn't define the logic of Tx execution depending on the error, I choose to end the Tx execution.
It avoids having a Tx that always restart even after a reboot.

PR #145, opened 2024-03-18
Add protocol version in peer handshake

musitdev:
I've added the exchange of the protocol version (u64=0) during the handshake.


tuommaki (approved):
LGTM :+1:

PR #148, opened 2024-03-18
Move rest of the locks to `SchedulerState`

tuommaki:
`Scheduler` had at least one more deadlock hiding in it's code. Most probably it was the `ProgramManager`. Moving both, the `mempool` and `program_manager` to `SchedulerState` doesn't reveal the deadlock anymore.


tuommaki:
Going with this to get a build for devnet operators still today.

@musitdev Please still check the diff and give me a shout if you think there are any issues with this.

musitdev:
I don't think that  putting together the mutex will solve it. It just changes the probability to see it. We'll see in production, but it can be very annoying because silent deadlock become visible when the load increases.
I think the sooner we remove the mutex by re-architecting, the better it will be.

PR #149, opened 2024-03-18
Fix workflow steps indexing condition

tuommaki:


crates/node/src/workflow/mod.rs
```diff
@@ -82,7 +82,7 @@ impl WorkflowEngine {
 
                 match workflow.steps.iter().position(|s| s.program == *prover) {
                     Some(proof_step_idx) => {
-                        if workflow.steps.len() <= proof_step_idx {
+                        if workflow.steps.len() <= (proof_step_idx + 1) {
```
tuommaki, line 85: Under stress test I faced a situation where `workflow.steps.len()` was `2` and `proof_step_idx` was `1`. This obviously then crashed on line 95...

tuommaki:
All green, bug pretty obvious, going...

PR #150, opened 2024-03-18
Adjust E2E test prover&verifier resource reqs

tuommaki:
Having less requirements for prover & verifier in E2E test, the more concurrency it allows when used for stress testing.

### DIFF
--- a/crates/tests/e2e-tests/src/main.rs
+++ b/crates/tests/e2e-tests/src/main.rs
@@ -2,6 +2,7 @@ use clap::Parser;
 use gevulot_node::{
     rpc_client::RpcClient,
     types::{
+        program::ResourceRequest,
         transaction::{Payload, ProgramData, ProgramMetadata, Workflow, WorkflowStep},
         Hash, Transaction,
     },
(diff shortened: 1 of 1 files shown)


tuommaki:
One can test the node locking pretty nicely with this:

```go
diff --git a/crates/tests/e2e-tests/src/main.rs b/crates/tests/e2e-tests/src/main.rs
index c16c285..30be08d 100644
--- a/crates/tests/e2e-tests/src/main.rs
+++ b/crates/tests/e2e-tests/src/main.rs
@@ -56,10 +56,11 @@ async fn main() -> Result<()> {
     .await
     .expect("deploy");
 
-    for nonce in 1..2 {
+    for nonce in 1..4000 {
         send_proving_task(&client, &key, nonce, &prover_hash, &verifier_hash)
             .await
             .expect("send proving task");
+        sleep(Duration::from_millis(350)).await;
     }
 
     sleep(Duration::from_secs(360)).await;
```

kylegranger (approved):
LGTM 👍

PR #151, opened 2024-03-18
Change tx root folder to txfiles

musitdev:
This version use the same folder for Virtfs and Tx proof input files. It avoids copying the files.

I change the AssetFile to only use the file path, and the path construction logic is done when the asset file is created.

VM_FILES_DIR is not implemented.

### DIFF
--- a/crates/node/src/types/file.rs
+++ b/crates/node/src/types/file.rs
@@ -212,7 +212,7 @@ impl AssetFile {
                 checksum,
             } => {
                 //verify the url is valide.
-                reqwest::Url::parse(file_url)?;
+                reqwest::Url::parse(&file_url)?;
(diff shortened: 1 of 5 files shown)


crates/node/src/types/file.rs
```diff
@@ -187,42 +194,53 @@ pub struct DbFile {
...
-                url,
+    pub fn new_from_program_data(data: &ProgramData, tx_hash: Hash) -> Result<Option<Self>> {
+        match data {
+            ProgramData::Input {
+                file_name,
+                file_url,
                 checksum,
-            },
+            } => {
+                //verify the url is valide.
```
tuommaki, line 214:
```suggestion
                // Verify the URL is valid.
```

tuommaki (approved):
LGTM :+1:

PR #153, opened 2024-03-19
Fix image filename

tuommaki:
When generating file name from a program deployment, use the `image_file_name` instead of program name / tag.


tuommaki:
This is a critical fix for the node to work at all if prover/verfier names are used. Tested on my node and it worked after the fix.

PR #154, opened 2024-03-19
Move VM VirtFs file folder to vmfiles

musitdev:
VM VirtFs workspace folder is now under vmfiles/<tx hash>/workspace.

Files are copied to and from between the node Tx and the VM VirtFs.

I put a cleanup of the workspace after Tx execution.


tuommaki (approved):
LGTM :+1:

PR #155, opened 2024-03-19
Introduce dummy lock to sequentialize P2P.connect()

tuommaki:
So far this doesn't seem to break anything. Testing more to see if the state keeps that way.


musitdev (approved):

tuommaki:
Node stayed up over an hour & still worked correctly. Let's merge and test more.

PR #157, opened 2024-03-20
P2p locking adjustments

tuommaki:


crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -758,6 +756,122 @@ mod tests {
...
+                )
+                .await
+                .expect(&format!("peer{i} connect p2p_beacon_node"));
+            tracing::debug!("peer{i} is connected to P2P beacon node");
+
+            peers.push(peer);
+        }
+
+        // Let the dust settle down a bit.
+        //tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
tuommaki, line 837: @musitdev in the current setting, if there's no wait / sleep between connects and transaction sends, this test locks up on my laptop. What happens is that those derivative new node connects are happening concurrently with the transaction sends and this causes the deadlock somehow.

If you uncomment this sleep and all the connects take place before the transaction sends, so far I couldn't reproduce the deadlock anymore.

tuommaki:
Most of the testing here is now finished and there are no further actions in this branch. Closing.

PR #161, opened 2024-03-21
Add timeout to QEMU QMP connect

tuommaki:
QEMU QMP connection can be blocking sometimes, which leads to whole scheduling loop getting halted. This change adds timeout to QMP connect so that there's maximum limit how long this can take time before the VM is given up and killed.

While at it, cleanup some leftovers from workspace volume refactoring and fix `vm_registry` maintenance on VM termination.


crates/node/src/vmm/qemu.rs
```diff
@@ -187,7 +189,6 @@ impl Provider for Qemu {
             tx_hash,
             program_id,
             workspace_volume_label,
-            //qmp: Arc::new(Mutex::new(qmp)),
```
tuommaki, line 190: Old leftovers.

crates/node/src/vmm/qemu.rs
```diff
@@ -224,7 +225,6 @@ impl Provider for Qemu {
             // Register 2 hard drives via SCSI
             .args(["-device", "virtio-scsi-pci,bus=pci.2,addr=0x0,id=scsi0"])
             .args(["-device", "scsi-hd,bus=scsi0.0,drive=hd0"])
-            //.args(["-device", "scsi-hd,bus=scsi0.0,drive=hd1"])
```
tuommaki, line 227: Same, leftovers.

crates/node/src/vmm/qemu.rs
```diff
@@ -242,15 +242,9 @@ impl Provider for Qemu {
                     &img_file.into_os_string().into_string().unwrap(),
                 ),
             ])
-            // WORKSPACE FILE
-            /*
-            .args([
-                "-drive",
-                &format!("file={},format=raw,if=none,id=hd1", &workspace_file),
-            ])*/
```
tuommaki, line 250: No need for this either :slightly_smiling_face:

crates/node/src/vmm/qemu.rs
```diff
@@ -309,19 +303,54 @@ impl Provider for Qemu {
...
+                                ))
+                                .and_then(|p| {
+                                    p.kill()?;
+                                    p.wait()
+                                })?;
+
+                            self.vm_registry.remove(&cid);
+                            self.cid_allocations.remove(&cid);
+                            return Err(eyre!("Failed to connect to QEMU QMP"));
+                        }
```
tuommaki, line 347: This is quite horrible but figured we need to do the cleanup and `&mut self` makes things slightly complicated here. :see_no_evil:

crates/node/src/vmm/qemu.rs
```diff
@@ -309,19 +303,54 @@ impl Provider for Qemu {
...
                     }
                 };
             }
             client.unwrap()
         };
 
-        qmp_client.system_reset().await?;
-
```
tuommaki, line 324: No need for extra VM reboot anymore, since we don't attach the disk during runtime via QMP. When I saw this, I got thinking that _maybe_ this could also have contributed to the issues with the gRPC-over-VSOCK :thinking: 

I bet that with right timing, the VM can probably _just_ start and connect, until it already gets smacked in the head for reboot. That would probably leave the server in very confused state.

crates/node/src/vmm/qemu.rs
```diff
@@ -350,6 +379,7 @@ impl Provider for Qemu {
 
             let cid = qemu_vm_handle.cid;
             self.release_cid(cid);
+            self.vm_registry.remove(&cid);
```
tuommaki, line 382: Whoops! I wonder how anything worked without this :laughing:

crates/node/src/scheduler/mod.rs
```diff
@@ -205,7 +205,15 @@ impl Scheduler {
...
                 let mut state = self.state.lock().await;
-                while let Some((tx_hash, program_id)) = state.pending_programs.pop_front() {
+
+                // Make a local copy of pending programs so that when some
+                // programs fail to start and are rescheduled, they won't
+                // cause an infinite loop here.
+                let mut current_pending_programs = state.pending_programs.clone();
+                while let Some((tx_hash, program_id)) = current_pending_programs.pop_front() {
+                    // Pop from the actual pending program queue as well.
+                    state.pending_programs.pop_front();
```
tuommaki, line 215: This should be correct, right?

The original queue is cloned, so that on this scheduler round only "old" entries are processed.

If some of the programs fail to start and needs rescheduling, it's added to the end of the `state.pending_programs` so that it gets processed on the next full round of scheduler loop.

crates/node/src/scheduler/mod.rs
```diff
@@ -220,10 +228,19 @@ impl Scheduler {
                             sleep(Duration::from_millis(500)).await;
 
                             // Return the popped program_id back to pending queue.
-                            state.pending_programs.push_front((tx_hash, program_id));
+                            state.pending_programs.push_back((tx_hash, program_id));
```
tuommaki, line 231: I feel like this old code here would have caused infinite loop, right? :grimacing:

crates/node/src/scheduler/mod.rs
```diff
@@ -299,6 +309,10 @@ impl Scheduler {
                         let program_id = &task.program_id.clone();
                         state.task_queue.remove(&task.tx);
                     }
+
+                    self.reschedule(&task).await?;
+                    tracing::warn!("task {} rescheduled: {}", task.id.to_string(), err);
+                    continue;
```
tuommaki, line 315: Now all errors are treated the same, unless the needed program is missing.

musitdev (approved):

PR #162, opened 2024-03-22
When `ReousceError` happens, sheduler should try to `pick_task`

ghostant-1017:
Assume we have 3 programs in `pending_programs`, we successfully start the 1st and 2nd programs, but fail to `start_program` 3rd, the logic should be continue to `pick_task`, right?

Please correct me if I was wrong.

### DIFF
--- a/crates/node/src/scheduler/mod.rs
+++ b/crates/node/src/scheduler/mod.rs
@@ -229,7 +229,7 @@ impl Scheduler {
 
                             // Return the popped program_id back to pending queue.
                             state.pending_programs.push_back((tx_hash, program_id));
-                            continue 'SCHEDULING_LOOP;
+                            break;
                         }
                         Err(e) => {
                             tracing::warn!(


tuommaki:
> Assume we have 3 programs in pending_programs, we successfully start the 1st and 2nd programs, but fail to start_program 3rd, the logic should be continue to pick_task, right?

No, I think.

There are two things here:
1. The original `pending_programs` queue has been cloned. Every element in it should be processed.
2. In the middle of the cloned `pending_programs` queue, there can be programs that have very high resource requirements, that don't fit the node _at the moment_, but later in the queue there might be less resource intensive programs that _do_ fit - those should be scheduled to run, before moving to next new task.

The scheduling policy here is ultimately a decision on priority vs. utilization. Currently we don't have proper real world experience from all different kinds of workloads, so we don't know what's the best approach in terms of throughput vs. fairness.

The current implementation does have a negative property in terms of favoring small tasks and possibly discriminating very large tasks, but currently I'm not sure if this is a bad thing. The thing is that given the circumstances in devnet, this reduces chances for annoying or malicious behavior.

PR #163, opened 2024-03-22
correct Mac build by adding a common crate to share common declarations

musitdev:
Add the common crate to avoid to build shim crate when building node::type::files.rs because of the share const WORKSPACE_PATH.

Start to put common data/function/struct in an independent crate.

@teempai, it builds on my Mac. You can test. Go in crates/cli and cargo run


tuommaki (approved):
LGTM :+1:

PR #164, opened 2024-03-22
Fix deadlock on task scheduling when VM start fails

tuommaki:
When VM startup fails, the task is rescheduled and after recent changes to scheduler state management, this structure caused deadlock.

Inline task rescheduling to avoid deadlock.

### DIFF
--- a/crates/node/src/scheduler/mod.rs
+++ b/crates/node/src/scheduler/mod.rs
@@ -310,7 +310,9 @@ impl Scheduler {
                         state.task_queue.remove(&task.tx);
                     }
 
-                    self.reschedule(&task).await?;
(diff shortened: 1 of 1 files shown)


crates/node/src/scheduler/mod.rs
```diff
@@ -310,7 +310,9 @@ impl Scheduler {
                         state.task_queue.remove(&task.tx);
                     }
 
-                    self.reschedule(&task).await?;
```
tuommaki, line 313: `self.state` is locked _here_, but `fn reschedule()` tried to lock it again, in order to push the task back to queue.

musitdev (approved):

PR #166, opened 2024-03-26
Add command to print public key of a keyfile

phy-chain:
For easy debugging, you can print the public key associated with a specific keyfile (useful for devnet especially to figure out which key you registered in case you generated several)

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -7,6 +7,7 @@ use gevulot_node::types::Hash;
 use gevulot_node::types::TransactionTree;
 use std::net::SocketAddr;
 use std::path::PathBuf;
+use libsecp256k1::PublicKey;
 
 #[derive(Parser, Debug)]
 #[clap(author = "Gevulot Team", version, about, long_about = None)]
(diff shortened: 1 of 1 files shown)


crates/cli/src/main.rs
```diff
@@ -192,6 +194,13 @@ async fn main() {
             ),
             Err(err) => println!("Error during key file creation:{err}"),
         },
+        ConfCommands::PrintPublicKey => match gevulot_cli::keyfile::read_key_file(&args.keyfile) {
```
tuommaki, line 197: `read_key_file()` returns `SecretKey`.

tuommaki (changes requested):
Current implementation would print the secret key.

Please use `PublicKey::from_secret_key()` to extract public key.

crates/cli/src/main.rs
phy-chain: Oh yeah, you're right, misread the code !
Should be fixed.

tuommaki:
Please address the `cargo fmt` errors :slightly_smiling_face:

tuommaki:
To get this through, I went and implemented this once and for all in #172. Closing here.

PR #167, opened 2024-03-27
add watchdog http server

musitdev:
Add a http server that return the codes:
 *  Alive => StatusCode::OK (200)
 * Graceful => StatusCode::NO_CONTENT (204)
 * Critical => StatusCode::SERVICE_UNAVAILABLE (503)

2 signals are implemented:
 * the scheduler loop doesn't run
 * the mempool has more than 0 Tx after pooling.

Add a timer to detect that the mempool pick a task is not called after a certain time.

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -129,11 +129,11 @@ pub struct Config {
 
     #[arg(
         long,
-        long_help = "Watchdog listen address",
-        env = "GEVULOT_WATCHDOG_LISTEN_ADDR",
+        long_help = "Healthcheck listen address",
+        env = "GEVULOT_HEALTHCHECK_LISTEN_ADDR",
         default_value = "127.0.0.1:8888"
     )]
-    pub http_watchdog_listen_addr: SocketAddr,
+    pub http_healthcheck_listen_addr: SocketAddr,
 }
 
 #[derive(Debug, Args)]
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -129,11 +129,11 @@ pub struct Config {
 
     #[arg(
         long,
-        long_help = "Healthcheck listen address",
-        env = "GEVULOT_HEALTHCHECK_LISTEN_ADDR",
+        long_help = "Watchdog listen address",
+        env = "GEVULOT_WATCHDOG_LISTEN_ADDR",
         default_value = "127.0.0.1:8888"
     )]
-    pub http_healthcheck_listen_addr: SocketAddr,
+    pub http_watchdog_listen_addr: SocketAddr,
 }
 
 #[derive(Debug, Args)]
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -126,6 +126,14 @@ pub struct Config {
 
     #[arg(long, long_help = "GPU PCI devices", env = "GEVULOT_GPU_DEVICES")]
     pub gpu_devices: Option<String>,
+
+    #[arg(
+        long,
+        long_help = "Watchdog listen address",
+        env = "GEVULOT_WATCHDOG_LISTEN_ADDR",
+        default_value = "127.0.0.1:8888"
+    )]
+    pub http_watchdog_listen_addr: SocketAddr,
 }
 
 #[derive(Debug, Args)]
--- a/crates/node/src/scheduler/mod.rs
+++ b/crates/node/src/scheduler/mod.rs
@@ -13,6 +13,7 @@ use crate::vmm::qemu::Qemu;
 use crate::vmm::vm_server::grpc;
 use crate::vmm::vm_server::TaskManager;
 use crate::vmm::vm_server::VMServer;
+use crate::watchdog::HealthCheckSignal;
 use crate::workflow::{WorkflowEngine, WorkflowError};
 use crate::{
     mempool::Mempool,
@@ -37,6 +38,7 @@ use std::{
 use systemstat::ByteSize;
 use systemstat::Platform;
 use systemstat::System;
+use tokio::sync::mpsc;
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::{
     sync::{Mutex, RwLock},
@@ -197,7 +199,7 @@ impl Scheduler {
         }
     }
 
-    pub async fn run(&self) -> Result<()> {
+    pub async fn run(&self, watchdog_sender: mpsc::Sender<HealthCheckSignal>) -> Result<()> {
         'SCHEDULING_LOOP: loop {
             // Reap terminated tasks & VMs.
             self.reap_zombies().await;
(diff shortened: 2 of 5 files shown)


crates/node/src/cli.rs
```diff
@@ -126,6 +126,14 @@ pub struct Config {
...
     #[arg(long, long_help = "GPU PCI devices", env = "GEVULOT_GPU_DEVICES")]
     pub gpu_devices: Option<String>,
+
+    #[arg(
+        long,
+        long_help = "Watchdog listen address",
+        env = "GEVULOT_WATCHDOG_LISTEN_ADDR",
+        default_value = "127.0.0.1:8888"
+    )]
+    pub http_watchdog_listen_addr: SocketAddr,
```
tuommaki, line 136:
```suggestion
        long_help = "Watchdog listen address",
        env = "GEVULOT_WATCHDOG_LISTEN_ADDR",
        default_value = "127.0.0.1:8888"
    )]
    pub http_watchdog_listen_addr: SocketAddr,
```
I'd maybe name this as a healthcheck for the end user. The end user doesn't care how it has been implemented, but rather what it can provide. Healthcheck is common name for an HTTP interface that provides liveness status from `/healthz`.
```suggestion
        long_help = "Healthcheck listen address",
        env = "GEVULOT_HEALTHCHECK_LISTEN_ADDR",
        default_value = "127.0.0.1:8888"
    )]
    pub http_healthcheck_listen_addr: SocketAddr,
```

crates/node/src/scheduler/mod.rs
```diff
@@ -245,12 +247,25 @@ impl Scheduler {
...
+            let (mut task, mempool_size) = match self.pick_task().await {
+                (Some(t), size) => {
                     tracing::debug!("task {}/{} scheduled for running", t.id, t.tx);
-                    t
+                    (t, size)
                 }
-                None => {
+                (None, size) => {
+                    if let Err(err) = watchdog_sender
+                        .send(HealthCheckSignal::SchedulerMempoolLen(0))
```
tuommaki, line 264: Double-check: Should the parameter here be `0` or `size` ?

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+use std::sync::Arc;
+use tokio::net::TcpListener;
+use tokio::select;
+use tokio::sync::mpsc;
+use tokio::sync::Mutex;
+use tokio::time::timeout;
+use tokio::time::Duration;
+
+pub const GRACEFULL_SIGNAL_COUNTER_LIMIT: usize = 20;
+pub const SCHEDULER_HEATH_SIGNAL_TIMEOUT_MILLI: Duration = Duration::from_millis(1000);
```
tuommaki, line 16:
```suggestion
pub const SCHEDULER_HEALTH_SIGNAL_TIMEOUT_MILLIS: Duration = Duration::from_millis(1000);
```

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+pub const NO_LOOP_DETECT_TIMEOUT_MILLI: Duration = Duration::from_millis(1000);
```
tuommaki, line 17:
```suggestion
pub const NO_LOOP_DETECT_TIMEOUT_MILLIS: Duration = Duration::from_millis(1000);
```

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+    watchdog_state: Arc<Mutex<WatchDogState>>,
+    mut scheduler_health_rx: mpsc::Receiver<HealthCheckSignal>,
+) {
+    let mut scheduler_state = (true, true, true);
+
+    // Define a timer to detect when the mempool pick_task is never call but the loop is working.
+    let mut no_loop_detect_timer = tokio::time::interval(no_loop_detect_timeout);
+    let mut see_scheduler_loop_ok = true;
+
+    let mut timout_counter = 0;
```
tuommaki, line 76:
```suggestion
    let mut timeout_counter = 0;
```

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+        }
+    });
+
+    serve_files(bind_addr, watchdog_state).await?;
+    Ok(scheduler_health_tx)
+}
+
+async fn run_watchdog(
+    gracefull_signal_counter_limit: usize,
+    scheduler_heath_signal_timeout: Duration,
```
tuommaki, line 65:
```suggestion
    scheduler_health_signal_timeout: Duration,
```

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+            .await;
+        }
+    });
+
+    serve_files(bind_addr, watchdog_state).await?;
+    Ok(scheduler_health_tx)
+}
+
+async fn run_watchdog(
+    gracefull_signal_counter_limit: usize,
```
tuommaki, line 64:
```suggestion
    graceful_signal_counter_limit: usize,
```

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+    scheduler_heath_signal_timeout: Duration,
+    no_loop_detect_timeout: Duration,
+    watchdog_state: Arc<Mutex<WatchDogState>>,
+    mut scheduler_health_rx: mpsc::Receiver<HealthCheckSignal>,
+) {
+    let mut scheduler_state = (true, true, true);
```
tuommaki, line 70: Maybe create a separate struct with named fields for this? I feel that it's a bit difficult to follow the code when these state fields are only referenced by the tuple index.

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+            }
+        }
+    }
+}
+
+// Start the local server and serve the watch dog requert.
+// WatchDogState::Alive => 200
+// WatchDogState::Graceful => 204
+// WatchDogState::Critical => 503
+async fn serve_files(
```
tuommaki, line 148:
```suggestion
async fn serve_healthcheck(
```

crates/node/src/watchdog/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+                match listener.accept().await {
+                    Ok((stream, _from)) => {
+                        let io = TokioIo::new(stream);
+                        tokio::task::spawn({
+                            let watchdog_state = watchdog_state.clone();
+                            async move {
+                                if let Err(err) = http1::Builder::new()
+                                    .serve_connection(
+                                        io,
+                                        service_fn(|_| {
```
tuommaki, line 173: Having specific handler for `/healthz` would be good as there might be need for separate `/ready` or in some cases just a super simple '/' to get dummy HTTP 200 OK if the process exists.

tuommaki (commented):
Couple suggestions for improving maintainability of this.

crates/node/src/scheduler/mod.rs
musitdev: Yes, good catch, it's size.

tuommaki (approved):
Build green, let's go! :rocket:

musitdev:
I've still some change. I repoen a new PR.

PR #168, opened 2024-04-01
rpc_server: `send_transaction` returns more specific error details

ghostant-1017:
Currently, the error returned by `send_transaction` cannot be traced to a specific cause, leaving the developer unable to locate where the problem occurred.


tuommaki:
Hi! I understand your point in this PR, but as the Gevulot Devnet is currently a permissioned service, for the security reasons we follow best practices and keep the error codes / messages as generic in the API and don't reveal any technical information in them.

You can run local node for development purposes and there you can see the actual reason in the logs and if you want, you can also augment your local node instance with debugging capabilities.

PR #171, opened 2024-04-02
Verify that the program id is present or wait for the deploy Tx

musitdev:
The same way Tx wait for their parent, this PR add the wait for the deploy Tx associated to the Tx program id if not present.
Only Run tx are tested for program id because proof and verify Tx depend on the Run. If the Run wait for the program id proof and verify Tx will wait for the run.

I synchronize the wait verification and the Tx save because some time Wait Tx verification can return several Txs and if it contains a Run Tx for example, the dependent proof that arrive after can be saved before because the batch of Tx take more time to save than one Tx verification and save.
I add the case in my test.

I declare a specific storage trait for the verification.

With the `wait_if_not_cached` function I try to share the same logic between the program dep and the parent dep verification but not sure, it adds because of the complexity to make it with async future.

I test it with 3 nodes connected.

### DIFF
--- a/crates/node/src/txvalidation/event.rs
+++ b/crates/node/src/txvalidation/event.rs
@@ -270,6 +270,7 @@ impl TxEvent<WaitTx> {
         ) -> impl Future<Output = eyre::Result<bool>> + 'a {
             storage.contains_program(*cache_key)
         }
+
         // Test  all Tx progam id because Run tx can have program from different deploy Tx.
         let mut new_tx: Option<Self> = Some(self);
         for program_id in run_tx_programs {
--- a/crates/node/src/txvalidation/event.rs
+++ b/crates/node/src/txvalidation/event.rs
@@ -326,6 +326,13 @@ impl TxEvent<WaitTx> {
         }
         // Only test for the first program, because all Tx program came from the same deploy tx.
         let parent = self.tx.payload.get_parent_tx().cloned();
+        tracing::warn!(
+            "manage_parent_dep parent {:?}",
+            parent
+                .iter()
+                .map(|hash| hash.to_string())
+                .collect::<Vec<_>>()
+        );
         let new_tx = self
             .wait_if_not_cached(parent_cache, parent.as_ref(), storage, query_db_for_tx)
             .await?;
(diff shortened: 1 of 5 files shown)


crates/node/src/txvalidation/event.rs
```diff
@@ -196,53 +198,145 @@ impl TxEvent<PropagateTx> {
...
+    ) -> Result<Vec<TxEvent<WaitTx>>, EventProcessError> {
+        // Verify that Tx associated program are present.
+        let run_tx_programs = self.tx.payload.get_run_programs_dep();
+        fn query_db_for_program<'a>(
+            cache_key: &'a Hash,
+            storage: &'a dyn ValidateStorage,
+        ) -> impl Future<Output = eyre::Result<bool>> + 'a {
+            storage.contains_program(*cache_key)
+        }
+        // Only test for the first program, because all Tx program came from the same deploy tx.
```
tuommaki, line 271: This is not actually true. Nothing restricts Run transaction from combining programs from different deployments (and that is fine).

tuommaki (commented):
LGTM otherwise, but the wait should ensure that all programs are present, not only the first one, because Run transaction can use programs from different deployments.

crates/node/src/txvalidation/event.rs
musitdev: Yes, I made a shortcut. I'll change it.

musitdev: I've updated the PR to track all program id for a Run Tx.

crates/node/src/txvalidation/event.rs
```diff
@@ -196,53 +198,171 @@ impl TxEvent<PropagateTx> {
...
+        ) -> impl Future<Output = eyre::Result<bool>> + 'a {
+            storage
+                .get_tx(cache_key)
+                .and_then(|res| async move { Ok(res.is_some()) })
+        }
+        // Only test for the first program, because all Tx program came from the same deploy tx.
+        let parent = self.tx.payload.get_parent_tx().cloned();
+        let new_tx = self
+            .wait_if_not_cached(parent_cache, parent.as_ref(), storage, query_db_for_tx)
+            .await?;
```
tuommaki, line 338: Should this also verify all the programs and not just the first one?

tuommaki (approved):
LGTM :+1:

PR #172, opened 2024-04-03
Add `show-public-key` command to `gevulot-cli`

tuommaki:
Occasionally it is useful to be able to display the public key from a keyfile (e.g. when verifying the allowlist entry). This change adds the command to `gevulot-cli` so that the full node binary is not needed for this functionality.


kylegranger (approved):
LGTM :rocket:

PR #174, opened 2024-04-04
Initial work towards observability

tuommaki:
This is the first step towards improving observability of Gevulot node.

This change adds some initial basic Prometheus metrics and a simple optional HTTP server that serves the metrics.

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -36,13 +36,16 @@ futures-util = { version = "0.3", features = [ "io" ], optional = true }
 gevulot-shim = { path = "../shim", default-features = false, optional = true }
 home = { version = "0.5", optional = true}
 http-body-util = { version = "0.1", optional = true }
-hyper = { version = "1", features = ["full"], optional = true }
+hyper = { version = "1.2", features = ["full"], optional = true }
 hyper-util = { version = "0.1", features = ["full"], optional = true }
 num_cpus = { version = "1.4.0", optional = true }
+lazy_static = { version = "1.4", optional = true }
 lru = "0.12.3"
 num-traits = { version = "0.2", optional = true }
 parking_lot = { version = "0.12", optional = true }
 pea2pea = { version = "0.48", optional = true }
+prometheus = { version = "0.13", optional = true }
+prometheus-hyper = { version = "0.1", optional = true }
 prost = { version = "0.11", optional = true }
 qapi = { version = "0.14", features = [ "qmp", "async-tokio-net" ], optional = true }
 snow = { version = "0.9", optional = true }
@@ -69,10 +72,13 @@ node-binary = [
   "http-body-util",
   "hyper",
   "hyper-util",
+  "lazy_static",
   "num_cpus",
   "num-traits",
   "parking_lot",
   "pea2pea",
+  "prometheus",
+  "prometheus-hyper",
   "prost",
   "qapi",
   "snow",
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -134,6 +134,13 @@ pub struct Config {
         default_value = "127.0.0.1:8888"
     )]
     pub http_healthcheck_listen_addr: SocketAddr,
+
+    #[arg(
+        long,
+        long_help = "Metrics server listen address",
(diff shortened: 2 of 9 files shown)


crates/node/Cargo.toml
```diff
@@ -36,13 +36,16 @@ futures-util = { version = "0.3", features = [ "io" ], optional = true }
 gevulot-shim = { path = "../shim", default-features = false, optional = true }
 home = { version = "0.5", optional = true}
 http-body-util = { version = "0.1", optional = true }
-hyper = { version = "1", features = ["full"], optional = true }
+hyper = { version = "1.2", features = ["full"], optional = true }
```
tuommaki, line 39: Upgraded while here. Not fully necessary, but figured that this wouldn't hurt either.

crates/node/src/cli.rs
```diff
@@ -134,6 +134,13 @@ pub struct Config {
         default_value = "127.0.0.1:8888"
     )]
     pub http_healthcheck_listen_addr: SocketAddr,
+
+    #[arg(
+        long,
+        long_help = "Metrics server listen address",
+        env = "GEVULOT_METRICS_LISTEN_ADDR"
+    )]
+    pub http_metrics_listen_addr: Option<SocketAddr>,
```
tuommaki, line 143: The metrics server is optional.

crates/node/src/main.rs
```diff
@@ -182,6 +183,15 @@ impl workflow::TransactionStore for storage::Database {
 async fn run(config: Arc<Config>) -> Result<()> {
     let node_key = read_node_key(&config.node_key_file)?;
 
+    // Register metrics counters.
+    metrics::register_metrics();
```
tuommaki, line 187: Metrics are always registered and gathered. Later on, I'll provide a separate PR that allows metrics export over P2P. The local HTTP server is optional and up to node operator whether it's enabled or not.

crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -49,6 +50,8 @@ pub struct P2P {
 
     // Send Tx to the process loop.
     tx_sender: TxEventSender<P2pSender>,
+
+    protocol_version: u64,
```
tuommaki, line 54: Although the value of this is constant (at the moment), I moved it to member variable so that its value can be shared / used across the code in uniform way and it is changed only from one place.

crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -105,6 +109,8 @@ impl P2P {
         instance.enable_writing().await;
         instance.enable_disconnect().await;
 
+        metrics::P2P_PROTOCOL_VERSION.set(instance.protocol_version as i64);
```
tuommaki, line 112: This will provide us the way to monitor protocol levels on the network. :sunglasses:

crates/node/src/rpc_server/mod.rs
tuommaki, line 1: I tried to read through the `jsonrpsee` documentation on middlewares, but I couldn't find a way to implement a solution similar to servlet filters where a middleware would have control over the whole lifecycle of request processing (which would be cool for unified response time measurement).

I do understand that the tower service design is built to allow this, but it requires quite a different kind of structure and I'm not sure if it plays well together with the RpcModule.

So, in the end I implemented only the request counting as middleware and took the shortcut with the response timing. This isn't a problem now and most probably ever as we have so small number of RPC methods anyway.

crates/node/src/scheduler/mod.rs
tuommaki, line 1: The tx execution time measurement is not 100% perfect here now. It mostly covers the "happy path", but I intentionally left some parts not measured as they would produce false results and if there are some other corner cases, we need to have a look at them as well separately.

The important thing here is that this change will provide us the interesting information and in those failure corner cases where the measurement doesn't log the time, it's ok as the time itself is not really interesting there but rather why things went so badly.

All in all, this is only the very beginning and things will improve a lot from here onwards so one gotta start from somewhere :wink:

musitdev (approved):

tuommaki:
Tested this on my node. Didn't observe any issues. Merging :rocket:

PR #175, opened 2024-04-05
Add support for P2P protocol v1 (when orig. is v0)

tuommaki:
This change makes next step in P2P protocol versioning support. While at it, this also takes the first step to add metrics into v1 protocol.

The prometheus metrics export is not part of this change yet, though.

The protocol versioning change here is built to be backwards compatible. Message versions default to v0, but have support for optional v1 compatibility as well. This change does not yet have support for protocol version negotiation between nodes, though it might be that this won't be even necessary, if nodes upgrade fast enough.

### DIFF
--- a/crates/node/src/metrics/mod.rs
+++ b/crates/node/src/metrics/mod.rs
@@ -113,3 +113,8 @@ pub(crate) async fn serve_metrics(bind_addr: SocketAddr) -> Result<()> {
 
     Ok(())
 }
+
+pub(crate) fn export_metrics() -> Vec<u8> {
+    // TODO: Implement :)
+    Vec::new()
+}
--- a/crates/node/src/networking/p2p/pea2pea.rs
+++ b/crates/node/src/networking/p2p/pea2pea.rs
@@ -52,6 +52,7 @@ pub struct P2P {
     tx_sender: TxEventSender<P2pSender>,
 
     protocol_version: u64,
+    node_resources: (u64, u64, u64),
 }
 
 impl Pea2Pea for P2P {
@@ -72,6 +73,7 @@ impl P2P {
         peer_http_port_list: Arc<tokio::sync::RwLock<HashMap<SocketAddr, Option<u16>>>>,
         tx_sender: TxEventSender<P2pSender>,
         propagate_tx_stream: impl Stream<Item = Transaction<Validated>> + std::marker::Send + 'static,
+        node_resources: (u64, u64, u64),
     ) -> Self {
         let config = Config {
             name: Some(name.into()),
@@ -101,6 +103,7 @@ impl P2P {
             nat_listen_addr,
             tx_sender,
             protocol_version: 0,
+            node_resources,
         };
 
         // Enable node functionalities.
@@ -118,7 +121,7 @@ impl P2P {
                 pin!(propagate_tx_stream);
                 while let Some(tx) = propagate_tx_stream.next().await {
                     let tx_hash = tx.hash;
-                    let msg = protocol::Message::V0(protocol::MessageV0::Transaction(tx));
+                    let msg = protocol::v0::Message::V0(protocol::v0::MessageV0::Transaction(tx));
                     let bs = match bincode::serialize(&msg) {
                         Ok(bs) => bs,
                         Err(err) => {
@@ -147,7 +150,7 @@ impl P2P {
         }
     }
 
-    async fn build_handshake_msg(&self) -> protocol::Handshake {
+    async fn build_handshake_msg(&self) -> protocol::internal::Handshake {
         let my_local_bind_addr = self.node.listening_addr().expect("p2p node listening_addr");
 
         // If NAT listen address hasn't been set, default 0.0.0.0:<port> is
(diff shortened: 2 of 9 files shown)


crates/node/src/metrics/mod.rs
```diff
@@ -113,3 +113,8 @@ pub(crate) async fn serve_metrics(bind_addr: SocketAddr) -> Result<()> {
 
     Ok(())
 }
+
+pub(crate) fn export_metrics() -> Vec<u8> {
+    // TODO: Implement :)
```
tuommaki, line 118: This will follow in separate PR.

crates/node/src/networking/p2p/pea2pea.rs
```diff
@@ -165,34 +168,55 @@ impl P2P {
...
+                    self.public_node_key,
+                    protocol::internal::DiagnosticsResponse::Metrics(data),
+                )
+            }
+        };
+
+        let bs = Bytes::from(
+            protocol::new_serialize_msg(self.protocol_version, resp)
+                .expect("diagnostics response serialization"),
+        );
```
tuommaki, line 219: Nodes never communicate between each other with diagnostics messages. Hence we don't need to support the original protocol here.

crates/node/src/networking/p2p/protocol/internal.rs
tuommaki, line 1: This module contains the node-internal message format. Versioned protocol messages are converted to / from this.

crates/node/src/networking/p2p/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
+use eyre::{eyre, Result};
+
+pub mod internal;
+pub mod v0;
+pub mod v1;
+
+pub fn serialize_handshake(msg: internal::Handshake) -> Result<Vec<u8>> {
+    let msg = v0::Handshake::from(msg);
+    bincode::serialize(&msg).map_err(|e| e.into())
+}
```
tuommaki, line 10: Present state.

crates/node/src/networking/p2p/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+        ver => return Err(eyre!("unknown protocol version: {ver}")),
+    }
+    .map(|mut bs| {
+        let mut data = Vec::with_capacity(bs.len() + 8);
+        data.append(&mut protocol_version.to_be_bytes().to_vec());
+        data.append(&mut bs);
+        data
+    })
+    .map_err(|e| e.into())
+}
```
tuommaki, line 25: Once this code is running on all nodes, this version will replace the original `serialize_handshake`.

crates/node/src/networking/p2p/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+    if bs.len() < 9 {
+        return Err(eyre!("invalid handshake message: too short"));
+    }
+
+    match u64::from_be_bytes(bs[0..8].try_into().expect("convert slice to array")) {
+        0 => v0::Handshake::parse(&bs[8..]),
+        1 => v1::Handshake::parse(&bs[8..]),
+        ver => Err(eyre!("unknown protocol version: {ver}")),
+    }
+}
```
tuommaki, line 42: `deserialize_handshake` is backwards compatible. First it tries the present state where individual messages are not prepended with version and only then it tries versioned approach.

crates/node/src/networking/p2p/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+    if bs.len() < 9 {
+        return Err(eyre!("invalid protocol message: too short"));
+    }
+
+    match u64::from_be_bytes(bs[0..8].try_into().expect("convert slice to array")) {
+        0 => v0::Message::parse(&bs[8..]),
+        1 => v1::Message::parse(&bs[8..]),
+        ver => Err(eyre!("unknown protocol version: {ver}")),
+    }
+}
```
tuommaki, line 79: Like with handshake, ordinary messages are implemented the same way.

crates/node/src/networking/p2p/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+            serialize_msg(internal::Message::Transaction(orig_tx.clone())).expect("serialize tx");
+
+        let msg: v1::Message = bincode::deserialize(bs.as_ref()).expect("deserialize v1 message");
+        if let v1::Message::V1(v1::MessageV1::Transaction(tx)) = msg {
+            println!("deserialized tx: {:?}", tx);
+            assert_eq!(orig_tx, tx);
+        } else {
+            panic!("test failed: Couldn't deserialize transaction correctly.");
+        }
+    }
```
tuommaki, line 105: Serialization / deserialization of transaction, between v0 and v1 works because the structure is 1:1 identical.

crates/node/src/networking/p2p/protocol/v1.rs
```diff
@@ -0,0 +1,183 @@
...
+        minor: u16,
+        patch: u16,
+        build: String,
+    },
+    Resources {
+        cpus: u64,
+        mem: u64,
+        gpus: u64,
+    },
+    Metrics(Vec<u8>),
```
tuommaki, line 87: This will hold encoded prometheus metrics from the node. With this, we can later on just add new metrics and that won't require changes to P2P wire-protocol.

crates/node/src/scheduler/mod.rs
tuommaki, line 1: I extracted the node computation resource configuration code into `resource_manager` module so that it can be reused both in `Scheduler`, but also in `main.rs / run()` for P2P diagnostics.

musitdev (approved):

tuommaki:
I've been running this on our devnet node for some time now and it seems to co-operate with other nodes just fine. It can now also serve the resource requirement requests so :ok_hand: 

Merging now, to proceed the next stage in testing.

PR #180, opened 2024-04-08
Add prometheus metrics export for P2P

tuommaki:
This change adds support for prometheus metrics export via P2P protocol. Metrics are exported with `TextEncoder` and compressed with `zstd` on default compression level (`3` at the implementation time).


tuommaki:
As I already tested this and this is relatively small change, I'm going with this now to get it on second stage of testing. Ping me when there is anything to consider :slightly_smiling_face:

PR #182, opened 2024-04-10
Add the ability to join networks to "seed mode" so that we can run a seed cluster

bhechinger:
To be able to reasonably run this in kubernetes we need to be able to run multiple seed pods.

This change has nodes run in `p2p-beacon` mode to try to join an existing network. This allows us two things.
- Running a seed cluster where there are multiple seed nodes
- Completely shut down the seed nodes being able to start them back up by bootstrapping off an existing network

At start the seed will attempt to contact the nodes listed in the discovery list. After enough attempts instead of failing it will fall back to bootstrapping a brand new network.

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -365,22 +365,28 @@ async fn p2p_beacon(config: P2PBeaconConfig) -> Result<()> {
     tracing::info!("listening for p2p at {}", p2p_addr);
 
     let mut connected_nodes = 0;
-    for addr in config.p2p_discovery_addrs.clone() {
-        tracing::info!("connecting to p2p peer {}", addr);
-        match addr.to_socket_addrs() {
-            Ok(mut socket_iter) => {
-                if let Some(peer) = socket_iter.next() {
-                    let (connected, fail) = p2p.connect(peer).await;
-                    connected_nodes += connected.len();
-                    if !fail.is_empty() {
-                        tracing::info!("Peer connection, fail to connect to these peers:{fail:?}");
+    let mut try_count = 0;
+
+    while connected_nodes == 0 && try_count < 5 {
+        for addr in config.p2p_discovery_addrs.clone() {
+            tracing::info!("connecting to p2p peer {}", addr);
+            match addr.to_socket_addrs() {
+                Ok(mut socket_iter) => {
+                    if let Some(peer) = socket_iter.next() {
+                        let (connected, fail) = p2p.connect(peer).await;
+                        connected_nodes += connected.len();
+                        if !fail.is_empty() {
+                            tracing::info!("Peer connection, fail to connect to these peers:{fail:?}");
+                        }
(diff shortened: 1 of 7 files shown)


musitdev (approved):
I think there are some modifications that came from my branch that I merge yesterday. Perhaps you'll need to rebase from main?

crates/node/src/main.rs
```diff
@@ -366,8 +375,52 @@ async fn p2p_beacon(config: P2PBeaconConfig) -> Result<()> {
...
+
+        tokio::task::spawn(async move {
+            if let Err(err) = http1::Builder::new()
+                .serve_connection(io, service_fn(ok))
+                .await
+            {
+                tracing::error!("Error serving connection: {:?}", err);
+            }
+        });
     }
```
musitdev, line 426: To avoid starting a new http server you can move the lines 243,244:
```
       let scheduler_watchdog_sender =
        watchdog::start_healthcheck(config.http_healthcheck_listen_addr).await?;
```
Just before the `if` for example line 239 and remove these.
It will activate the health check watchdog for all type of node. The scheduler verification will fail, but the / entry point will be ok.
I was thinking it will log an error. If it's too annoying, keep the http server you add, and I'll remove it when I'll change the watchdog to activate per service.

bhechinger:
> I think there are some modifications that came from my branch that I merge yesterday. Perhaps you'll need to rebase from main?

I had already re-based but it mangled the formatting. I just needed to re-run `cargo fmt`

musitdev (approved):
Perhaps you can test the watchdog modification to see?

crates/node/src/main.rs
bhechinger: I wasn't sure if it was safe to re-use that so I just whipped up something quick to get me past the issue I had (without the readiness check kubernetes won't wait to start the next statefulset pod). I'm ok with ignoring any errors it logs for now so fix it when you get around to it.

flake.nix
```diff
@@ -0,0 +1,34 @@
...
+          devShells.default = let p = pkgs; in
+            pkgs.mkShell {
+              buildInputs =
+              [
+                p.git
+                p.opentofu
+                p.google-cloud-sdk-gce
+                (p.google-cloud-sdk.withExtraComponents [p.google-cloud-sdk.components.gke-gcloud-auth-plugin])
+                p.k9s
+                p.kube-capacity
```
tuommaki, line 27: These shouldn't be needed with the node directly, right?

tuommaki (approved):
I'm fine with the PR in general, but maybe Nix files could be extracted into separate PR and checked for dependencies?

bhechinger:
> I'm fine with the PR in general, but maybe Nix files could be extracted into separate PR and checked for dependencies?

Oh shoot, I missed this message before merging. Let's catch up on slack about how to handle this.

PR #184, opened 2024-04-10
removed packages not needed to build the node on nixos

bhechinger:
Forgot to clean up the flake and remove unneeded packages.


tuommaki (approved):
:+1:

PR #185, opened 2024-04-10
Just a loop wasn't enough, it goes too fast

bhechinger:
Just doing a loop to retry connecting to a potentially existing cluster was very unreliable. The loop goes too fast.

This adds a 1 second delay to the loop so the first pod/service have time to become available for the second one to join.

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -397,6 +397,8 @@ async fn p2p_beacon(config: P2PBeaconConfig) -> Result<()> {
                     tracing::error!("failed to resolve {}: {}", addr, err);
                 }
             }
+
+            tokio::time::sleep(Duration::from_secs(1)).await;
         }
 
         try_count += 1;


tuommaki (approved):
:+1:

crates/node/src/main.rs
```diff
@@ -388,6 +388,8 @@ async fn p2p_beacon(config: P2PBeaconConfig) -> Result<()> {
                     tracing::error!("failed to resolve {}: {}", addr, err);
                 }
             }
+
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
musitdev, line 392: You put the sleep in the main loop. It means that the loop wait if the connection succeeds. For example if you define 10 seeds that are ok, it will take 10s. Perhaps you want to connect fast if the connection fails to find a valid seed fast. If yes, you should put the sleep in the Error case (`Err(err) => {` / line 387 and `if !fail.is_empty() {` / line 380).

bhechinger: You're absolutely right on the placement of the delay. That should be in the outer while loop, not the inner for loop.

Connecting fast isn't really a huge deal, so I'm not sure it's worth the extra logic. Seed cluster bootstrap isn't something that really needs to be super fast. A few extra seconds delay (even 10+) isn't really a problem.

If you disagree I can certainly add some extra logic to this, however.

musitdev: Ok It's fine. It was just to show you the point to see its validity.

musitdev (approved):

crates/node/src/main.rs
bhechinger: I certainly appreciate that!

PR #193, opened 2024-05-07
feat:  Add request_timeout to rcp_client

0x5459:
Some RPC requests cannot complete the response within the default 60 seconds. Therefore, I added a argument `rpc_timeout` to customize the timeout.

### DIFF
--- a/crates/node/src/rpc_client/mod.rs
+++ b/crates/node/src/rpc_client/mod.rs
@@ -9,17 +9,46 @@ use jsonrpsee::{
     http_client::{HttpClient, HttpClientBuilder},
 };
 use std::error::Error;
+use std::time::Duration;
+
+/// A RPC client builder for connecting to the Gevulot network
+#[derive(Debug)]
+pub struct RpcClientBuilder {
+    request_timeout: Duration,
+}
+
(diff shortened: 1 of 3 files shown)


crates/node/src/rpc_client/mod.rs
```diff
@@ -9,17 +9,46 @@ use jsonrpsee::{
     http_client::{HttpClient, HttpClientBuilder},
 };
 use std::error::Error;
+use std::time::Duration;
+
+/// A RPC client builder for connecting to the Gevulot network
+#[derive(Debug)]
+pub struct RpcClientBilder {
```
tuommaki, line 16:
```suggestion
pub struct RpcClientBuilder {
```
Small typo here. Please fix it across the whole PR. :slightly_smiling_face:

tuommaki (changes requested):
Please fix the typo in the builder type name, otherwise this looks good.

crates/node/src/rpc_client/mod.rs
0x5459: fixed.

tuommaki (approved):

PR #195, opened 2024-05-22
Fix tempfile naming in download manager

tuommaki:
When multiple files with same name, but different extension are submitted for download concurrently, there was a conflict with the filename.

This change fixes by adding ".tmp" to filename instead of replacing the whole extension.


trusch (changes requested):
Lgtm except that the serving logic needs to be adapted. Your changes ADD .tmp as an extension but in the serving function the check that looks for temp files sets the extension, overwriting the existing one.

tuommaki:
>  Your changes ADD .tmp as an extension but in the serving function the check that looks for temp files sets the extension, overwriting the existing one.

Whoops! That was a **very** good catch! Thanks! Extracted the functionality and used it in both.

trusch (approved):

PR #196, opened 2024-06-04
fix(shim): remove incorrect path

0x5459:
I'm trying to fix the workspace path in the `shim`. Perhaps I'm wrong.

@tuommaki can help to review it when you are free?


tuommaki:
Oh boy! Your fix definitely seems correct. Good catch! Thank You!

tuommaki (approved):

PR #198, opened 2024-06-16
fix: dynamic QEMU path resolution

vlopes11:
Prior to this commit, it is assumed that the QEMU binary resides within the canonical Linux path. However, certain systems like NixOS may place QEMU in a local user folder instead of the global `/usr/bin` path, causing an error when attempting to locate the QEMU binary.

This change incorporates a `which` command, which searches through the current `PATH` environment variable and identifies the location of the QEMU binary.

### DIFF
--- a/crates/node/src/rpc_server/mod.rs
+++ b/crates/node/src/rpc_server/mod.rs
@@ -495,6 +495,7 @@ mod tests {
             http_download_port: 0,
             http_healthcheck_listen_addr: "127.0.0.1:8888".parse().unwrap(),
             http_metrics_listen_addr: None,
+            qemu_path: None,
         });
 
         let db = Arc::new(Database::new(&cfg.db_url).await.unwrap());


crates/node/src/vmm/qemu.rs
```diff
@@ -189,7 +189,10 @@ impl Provider for Qemu {
 
         // Update the child process field.
         let qemu_vm_handle = &mut self.vm_registry.get_mut(&cid).unwrap();
-        let mut cmd = Command::new("/usr/bin/qemu-system-x86_64");
+
+        // Resolve the QEMU system command
+        let qemu_path = which::which("qemu-system-x86_64")?;
+        let mut cmd = Command::new(qemu_path);
```
tuommaki, line 195: Instead of resolving the path on every VM startup, could this be done when constructing the QEMU provider at `new()` ?

tuommaki (approved):
That works! :ok_hand:

PR #201, opened 2024-06-28
Fix flag param handling in gevulot-cli

tuommaki:


koxu1996:
This PR should not go unnoticed, as it introduced important workaround for passing single value args.

### My case study

I wanted to pass `foo -- bar` args, so I used the following tasks' `cmd_args` when calling `gevulot-cli exec`

```
[{ "name": "foo", "value": "" }, { "name": "", "value": "--" }, { "name": "bar", "value": "" }]
```

Here is the request that was sent with CLI - [request-1.json](https://github.com/user-attachments/files/16781373/request-1.json):

```
     ...
              "args": ["foo", "", "", "--", "bar", ""],
     ...
```

I was not expecting those empty strings :crying_cat_face:.

The solution was to update the CLI tool with `cargo install --git https://github.com/gevulotnetwork/gevulot.git gevulot-cli` (from `#bbe20d41` to `#700d759a`).

Request after update - [request-2.json](https://github.com/user-attachments/files/16781418/request-2.json):

```
     ...
              "args": ["foo", "--", "bar"],
     ...
```

Now everything works as expected!

PR #202, opened 2024-07-01
Make `vsock` optional feature for `shim`

tuommaki:
When `shim` library is compiled without `vsock` feature, it defaults to file based approach for loading `Task` and writing `TaskResult`.

It uses `/workspace/task.json` for input and
`/workspace/task_result.json` for output.

### DIFF
--- a/crates/shim/src/lib.rs
+++ b/crates/shim/src/lib.rs
@@ -285,8 +285,11 @@ pub fn run(callback: impl Fn(Task) -> Result<TaskResult>) -> Result<()> {
     let task: Task = serde_json::from_reader(file)?;
 
     let result = callback(task).map_err(|e| e.to_string());
-    let mut result_file =
-        File::create_new(PathBuf::from(WORKSPACE_PATH).join(TASK_RESULT_FILE_NAME))?;
+    let mut result_file = File::options()
+        .read(true)
+        .write(true)
+        .create_new(true)
+        .open(PathBuf::from(WORKSPACE_PATH).join(TASK_RESULT_FILE_NAME))?;
     serde_json::to_writer(&mut result_file, &result)?;
     result_file.flush()?;
 


crates/shim/src/lib.rs
```diff
@@ -239,3 +261,34 @@ pub fn run(callback: impl Fn(Task) -> Result<TaskResult>) -> Result<()> {
...
+
+        sleep(Duration::from_secs(1));
+    }
+
+    let file = File::open(PathBuf::from(WORKSPACE_PATH).join(TASK_FILE_NAME))?;
+    let task: Task = serde_json::from_reader(file)?;
+
+    let result = callback(task).map_err(|e| e.to_string());
+    let mut result_file =
+        File::create_new(PathBuf::from(WORKSPACE_PATH).join(TASK_RESULT_FILE_NAME))?;
```
koxu1996, line 289: While trying to integrate Aztec's ACVM - [based](https://github.com/AztecProtocol/aztec-packages/blob/master/noir/noir-repo/rust-toolchain.toml) on stable Rust `v1.74.1` - I got the following error:

```
error[E0658]: use of unstable library feature 'file_create_new'
   --> /home/andrew/.cargo/git/checkouts/gevulot-cd67166d40edf61d/ad0cd45/crates/shim/src/lib.rs:289:9
    |
289 |         File::create_new(PathBuf::from(WORKSPACE_PATH).join(TASK_RESULT_FILE_NAME))?;
    |         ^^^^^^^^^^^^^^^^
    |
    = note: see issue #105135 <https://github.com/rust-lang/rust/issues/105135> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `gevulot-shim` (lib) due to previous error
```

The issue is that `Flie::create_new` got stabilized quite recently - it is [available](https://github.com/rust-lang/rust/blob/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/std/src/fs.rs#L436-L439) since Rust 1.77, which was released ~5 months ago.

@tuommaki What about replacing this line with `OpenOptions::new().read(true).write(true).create_new(true).open(path.as_ref())`, so we can support older projects?

tuommaki: Yeah, good shout! I actually hit this problem during the weekend as well so changed it now :+1:

aleasims:
I guess some changes in node are required? Otherwise how the file `tasks.json` will appear in [newly created `workspace`](https://github.com/gevulotnetwork/gevulot/blob/main/crates/node/src/vmm/qemu.rs#L201-L207) directory?

tuommaki:
> I guess some changes in node are required? Otherwise how the file tasks.json will appear in [newly created workspace](https://github.com/gevulotnetwork/gevulot/blob/main/crates/node/src/vmm/qemu.rs#L201-L207) directory?

Changes would be required for the devnet node, yes, but I don't think it is necesarily justified to change that anymore. The file based approach is rather used for local development & testing and once the VM works locally, then it can be re-build against the original vsock based shim and deployed to devnet. The API is 1:1 compatible so it's only a re-build to switch the deployment target.

koxu1996 (approved):
LGTM :rocket:

Potential improvement for the future: by introducing automatic fallback to the local executor environment, we could avoid rebuilding binary.

PR #206, opened 2024-08-28
Parse json args as vector of strings

koxu1996:
Asking for args in `key=value` format is a mistake, as it is often necessary to pass single values for CLI execution. There was workaround #201, but I believe we should just use a vector of strings. Not only it's simpler and more intuitive, but it also supports weird case when user wants to pass empty string as an argument.

This is **breaking change** of CLI, but resolves all the issues with passing args. I will update documentation accordingly, once this PR is merged.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -28,21 +28,9 @@ pub async fn calculate_hash_command(file_path: &PathBuf) -> BoxResult<String> {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct JsonCmdArgs {
-    pub name: String,
-    pub value: String,
-}
-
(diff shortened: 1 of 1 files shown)


tuommaki:
> Asking for args in key=value format is a mistake, as it is often necessary to pass single values for CLI execution.

I :100: agree with you on this, but I'd rather not merge this in at this point. There's more than our documentation (e.g. blog articles, TG support / examples etc.) with the old format and I don't want to break them at this point when the devnet as-is won't be live too long anymore. We must ensure that the future APIs handle arguments correctly (which they do atm. AFAIK) and just put up with this unfortunate `key=value` brainfart.

koxu1996:
What about supporting both formats and printing deprecation warning when `key=value` is used?

